PR #1, opened 2022-02-18
[GLT-3640] initial project setup

djcooke:
Included test pages/scripts/endpoints for demonstration purposes. It'll be removed later when there is useful stuff in its place.

@asofmosq and @avarsava: I'd mainly appreciate your feedback on the Node/TypeScript/Webpack setup since you're more experienced with all of that than I. Any other feedback welcome too of course!

Anyone: Let me know if you'd like more/better documentation for any part of this


README.md
```diff
@@ -1,2 +1,27 @@
...
+Run server on default port (8080): `mvn spring-boot:run`
+
+Run server on different port: `mvn spring-boot:run -Dspring-boot.run.jvmArguments="-Dserver.port=8081"`
+
+## Monitoring Endpoints
+
+The following endpoints are available on the deployed webapp for monitoring purposes.
+
+* Health status: `/actuator/prometheus`
+* Prometheus metrics: `/actuator/prometheus`
```
callunity, line 27: Are these intended to be the same endpoint?

callunity, line 27: Why `/prometheus` and not `/metrics` like our other systems expose?

djcooke: This is the default endpoint Spring Boot uses. It's possible to change it via properties, but I figured keep it as simple as possible. I don't think it really matters for the Prometheus-side config?

djcooke: Nope! 😆  First should be `/actuator/health`. I'll fix that

callunity: I'm sure it's possible to change the prometheus config to go looking at this endpoint, but then humans have to remember it too - but maybe I'm the odd one out in actually searching for something on the /metrics pages on occasion?

djcooke: That's a fair point. I like having it under `/actuator/` because there are other actuator endpoints that we may choose to enable in the future, and having them all under the same path makes the security config simpler. I'll move `/actuator/prometheus` to `/metrics` as a special case though

src/main/java/ca/on/oicr/gsi/dimsum/Beans.java
```diff
@@ -0,0 +1,15 @@
+package ca.on.oicr.gsi.dimsum;
+
+import nz.net.ultraq.thymeleaf.layoutdialect.LayoutDialect;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class Beans {
```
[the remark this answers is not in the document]
djcooke: Beans! 😆 These are objects that are eligible for Spring autowiring. The existing one is required to support Thymeleaf Layout Dialect - during startup, the bean is created and autowired into the other Thymeleaf beans that are already being created automatically by Sprint Boot autoconfig magic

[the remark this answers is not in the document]
avarsava: Is there a design doc for dimsum? My experience with Spring autowiring is 95% 'headache' and I'm wary that a skeleton PR is already relying on "magic", but if there are requirements that can only be fulfilled in this case with Spring I'd like to read about it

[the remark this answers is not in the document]
djcooke: Spring Boot is an extremely popular framework with a ton of its own documentation. Dependency injection (autowiring), and autoconfiguration (based on libraries existing in the classpath) are a couple of its main features and the reasons for using it. I can mention these in the developer guide and link to the relevant Spring documentation if that's useful? I'm also happy to help if you have any questions or want to better understand autowiring.

There is a design document on the wiki and a developer guide included in this PR. If there's anything in particular that you think would be helpful to have documented somewhere, please let me know.

README.md
djcooke: Prometheus metrics are at `/metrics` now. This required changing the base path for all actuator endpoints, so now we have to specify a full path for any that we enable, but that's obvious enough if/when we enable others since the config is in the same place. I disabled the health endpoint since we don't have a use for it - Prometheus will be used to monitor health

mlaszloffy (approved):

package.json
```diff
@@ -0,0 +1,21 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "jquery": "^3.6.0"
+  },
+  "devDependencies": {
+    "@types/jquery": "^3.5.13",
+    "ts-loader": "^9.2.6",
+    "typescript": "^4.5.5",
```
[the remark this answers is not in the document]
djcooke: These are all used for compilation, which is part of development. To my (limited) knowledge, only the generated JavaScript is used at runtime, which doesn't require any of these libraries. The test page shows that it does in fact all work as is.

Is there something I'm missing?

callunity (approved):

asofmosq (approved):

PR #2, opened 2022-05-05
[GLT-3660] added SAML authentication

djcooke:


example-application.properties
```diff
@@ -0,0 +1,11 @@
+# This file should be copied and filled in to provide instance-specific configuration
+
```
mlaszloffy, line 2: server.port=8081

src/main/java/ca/on/oicr/gsi/dimsum/SecurityConfiguration.java
```diff
@@ -0,0 +1,91 @@
...
+@Profile("!noauth")
+public class SecurityConfiguration {
+
+  private static final String LOGIN_URL = "/login";
+
+  @Bean
+  public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
+    http
+        .authorizeHttpRequests(auth -> auth
+            .antMatchers("/favicon.ico").permitAll()
```
mlaszloffy, line 36: /metrics

mlaszloffy (approved):

example-application.properties
djcooke: Not required. If you're not already running Tomcat (or anything else on port 8080) then the default is fine. For production, we'll most likely put NGINX in front of it, so the default port is fine.

src/main/java/ca/on/oicr/gsi/dimsum/SecurityConfiguration.java
djcooke: added

asofmosq (approved):

PR #5, opened 2022-05-27
[GLT-3670] added loader and service for QC Gate data

djcooke:


djcooke:
Need to add `datadirectory` property to example `application.properties` file, but that file is created in #2. Will update this PR once that one is merged

PR #13, opened 2022-06-14
[GLT-3698] added JIRA link property for bug reports

Arcslogger:
can be found as `jira.link`, currently blank as we don't have a full URL

### DIFF
--- a/src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/CommonModelAttributeProvider.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/CommonModelAttributeProvider.java
@@ -14,6 +14,8 @@ public class CommonModelAttributeProvider {
 
   @Value("${build.version}")
   private String buildVersion;
+  @Value("${bugreport.url}")
+  private String bugReportUrl;
 
   @ModelAttribute("dataAgeMinutes")
   public long getDataAgeMinutes() {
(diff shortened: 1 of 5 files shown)


djcooke (commented):
`jira.link` is going to be confusing later when we have more JIRA integration and lots of links. There's also no reason this needs to be JIRA at all - we could decide to change the link to some other system later on. Please rename to `bugreport.url` or something similar.

You'll have to provide a value for the property in `test.properties` for the test context to start up successfully. Can use `https://www.example.com` or something

src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/CommonModelAttributeProvider.java
```diff
@@ -2,18 +2,27 @@
...
 
-@ControllerAdvice(basePackages = {"ca.on.oicr.gsi.dimsum.controller.mvc"})
+@ControllerAdvice(basePackages = { "ca.on.oicr.gsi.dimsum.controller.mvc" })
 public class CommonModelAttributeProvider {
 
   @Autowired
   private CaseService caseService;
 
+  @Value("${bugreport.url}")
+  private String bugReportURL;
```
djcooke, line 16: Please rename to `bugReportUrl`. There's not really a clear standard for this; just want to be consistent with other naming that includes acronyms/abbreviations (e.g. `RequisitionQc`, `Sample.id`)

Same in model attribute name

djcooke (approved):

mlaszloffy (approved):

PR #17, opened 2022-06-17
add vscode workspace config to git; configure formatters

djcooke:


dev_guide.md
```diff
@@ -2,19 +2,27 @@
...
 [README](/README.md) for more information on the external properties file.
 
+## Code Formatting
+
+Dim Sum has formatters configured for Java, HTML, and TypeScript in VSCode. For these to work, you
+must have the following extensions installed:
+
+- Language Support for Java(TM) by Red Hat
+- Prettier - Code Formatter
+
```
Arcslogger, line 25: Should `HTML Language Features` not be on here as well?

djcooke: I'm pretty sure that's built in and can't be installed/uninstalled?

djcooke:
Found that you can add workspace "recommended extensions" that vscode will automatically suggest to install, so I added those

Arcslogger (approved):

PR #21, opened 2022-06-27
[GLT-3708] styled table elements

Arcslogger:


src/main/resources/templates/test.html
```diff
@@ -11,7 +11,7 @@
   <div layout:fragment="content">
     <h1>Test Page</h1>
 
-    <div id="tableContainer"></div>
+    <div id="tableContainer" class="rounded-lg border-grey-200 border-2  overflow-auto"></div>
```
djcooke, line 14: * Borders should be applied to the table rather than its container
* `overflow-auto` should go on a div that contains **only** the table, as this container will also contain other elements (paging controls, etc.) that are outside of the table and should not scroll with the table

ts/util/html-utils.ts
```diff
@@ -0,0 +1,23 @@
+// adds a header cell to a table header
+export function styleColumnHeader(th: HTMLTableCellElement, header: string) {
+  th.className = "p-4 text-white font-semibold bg-grey-300";
+  th.appendChild(document.createTextNode(header));
+}
+
+// adds link-styled text to a div
+export function styleNameLink(div: HTMLDivElement, name: string) {
```
djcooke, line 8: Change this to `addLink(container: HTMLElement, text: string, url: string)`
* obvious that it's adding to the DOM
* can be used to append to any element, rather than only a `div`
* fully creates the link
* styling should be on the link (`<a>` tag) itself - not on the container, which may contain other elements in addition to this link
* Hard-code callers to link to "#" for now just for display purposes. We'll update it with the correct URLs later

tailwind.config.js
```diff
@@ -1,48 +1,49 @@
 module.exports = {
   content: [
     // files that Tailwind scans for
-    'src/main/resources/templates/*.html',
-    'ts/*.ts'
+    "src/main/resources/templates/*.html",
+    "ts/*.ts",
```
djcooke, line 5: You had changed this to `ts/**/*.ts` previously, to include subdirectories, but it looks like you reversed that change. Please change it back again. This is likely why Tailwind didn't pick up one of your classes

ts/util/html-utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  a.className = "text-green-200 font-bold hover:underline";
+  a.innerHTML = text;
+  container.appendChild(a);
+}
+
+// updates an empty col with corresponding style
+export function styleEmptyColumn(td: HTMLTableCellElement) {
+  td.className = "text-grey-300 bg-grey-200 border-grey-200 border-t-2";
+  td.appendChild(document.createTextNode("N/A"));
+}
```
djcooke, line 29: This function is doing a lot more than it should

* Setting the contents ("N/A") will normally be handled by the getParent/ChildContents functions. It's only for cases where a table has child columns and a parent object has no children that the table builder should set "N/A" itself. Move adding the text node back to the table builder
* "styleEmptyColumn" is unclear - "N/A" is "empty"? It's also only styling a cell, not an entire column. Rename to "styleNotApplicableCell"
* top cell borders (`border-grey-200 border-t-2`) are added in 4 different places. Remove them from here and instead, add a "addCell(tr)" function to table builder that calls tr.insertCell, sets these 2 classes, and returns the cell. Then change table-builder to call that function wherever it's currently calling tr.insertCell. `p-4` should probably be included for all cells too

ts/util/html-utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// updates an empty row with corresponding style
+export function styleNoDataRow(cell: HTMLTableCellElement) {
+  cell.className =
+    "text-black bg-grey-100 font-bold border-grey-200 border-t-2";
+  cell.appendChild(document.createTextNode("NO DATA"));
+}
```
djcooke, line 36: This will never be used anywhere else. Do it all in table-builder.addNoDataRow and remove this utility function

ts/util/table-builder.ts
```diff
@@ -137,22 +143,27 @@ export class TableBuilder<ParentType, ChildType> {
     const row = tbody.insertRow();
     const cell = row.insertCell();
     cell.colSpan = this.definition.columns.length;
-    const text = document.createTextNode("No data");
-    cell.appendChild(text);
+    cell.className =
+      "text-black p-4 bg-grey-100 font-bold border-grey-200 border-t-2";
+    cell.appendChild(document.createTextNode("NO DATA"));
```
djcooke, line 148: Use `addCell` so you don't have to repeat the common classes here

ts/util/table-builder.ts
```diff
@@ -137,22 +143,27 @@ export class TableBuilder<ParentType, ChildType> {
...
     if (!column.addParentContents) {
       throw new Error(
         `Column "${column.title}" specified as parent, but doesn't define addParentContents`
       );
     }
-    const td = tr.insertCell();
+    const td = addCell(tr);
+    if (index > 0) {
+      td.classList.add("border-l-2");
+    }
```
djcooke, line 166: Add `index` parameter to `addCell` and handle border in there instead, so it's consistent with `addColumnHeader`

djcooke (approved):

PR #26, opened 2022-07-04
[GLT-3717] Added Miso and Dashi icons

Arcslogger:


ts/util/html-utils.ts
```diff
@@ -33,6 +30,16 @@ export function addLink(container: HTMLElement, text: string, url: string) {
   container.appendChild(a);
 }
 
+export function addMisoIcon(container: HTMLElement, url: string) {
+  const a = document.createElement("a");
+  a.setAttribute("href", url);
+  const img = document.createElement("img");
+  img.src = "/img/miso_logo.svg";
+  img.alt = "link to MISO page";
```
djcooke, line 38: Please change alt text to "View in MISO"

ts/util/html-utils.ts
```diff
@@ -6,21 +6,18 @@ export function addColumnHeader(
...
 // adds a cell to a table row
 export function addCell(tr: HTMLTableRowElement, index: number) {
   const td = tr.insertCell();
-  td.className = "p-4 border-grey-200 border-t-1 text-left align-text-top";
-  if (index > 0) {
-    td.classList.add("border-l-1");
-  }
+  td.className =
+    "p-4 border-grey-200 border-t-1 text-left align-text-top" +
+    (index > 0 ? " border-l-1" : "");
```
djcooke, line 20: I think this (and similar above) was cleaner before, but it's fine either way

djcooke (approved):

PR #27, opened 2022-07-04
styling fixups

Arcslogger:
white space outside of the header logo is no longer a clickable link hot spot

### DIFF
--- a/ts/util/html-utils.ts
+++ b/ts/util/html-utils.ts
@@ -43,6 +43,7 @@ const highlightClasses: Record<CellStatus, string[]> = {
   error: ["bg-red"],
 };
 
+// shade cell backgrounds
 export function shadeElement(element: HTMLElement, status?: CellStatus | null) {
   if (status) {
     element.classList.add(...highlightClasses[status]);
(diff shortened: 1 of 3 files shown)


ts/util/html-utils.ts
```diff
@@ -51,6 +52,7 @@ export function shadeElement(element: HTMLElement, status?: CellStatus | null) {
 
 export type TextStyle = "error";
 
+// styles cell text to match cell shading
```
djcooke, line 55: Not necessarily related to cells or cell shading. We can define different text styles here that may or may not go with any particular shading, and may also use the error style in other places without shading

ts/util/html-utils.ts
```diff
@@ -51,6 +52,7 @@ export function shadeElement(element: HTMLElement, status?: CellStatus | null) {
 
 export type TextStyle = "error";
 
+// styles cell text to match appropriate text style
```
djcooke, line 55: Can apply to any text container, not just cells

djcooke (approved):

PR #28, opened 2022-07-04
[GLT-3720] added sorting controls to table builder

djcooke:
We may or may not want to change this to record the user's choice at some point, but for now it'll just reset to the default on page load, but not when changing filters, etc.

Styling not included in this PR; will be fixed up later

### DIFF
--- a/ts/util/table-builder.ts
+++ b/ts/util/table-builder.ts
@@ -3,11 +3,15 @@ import {
   addCell,
   addColumnHeader,
   CellStatus,
+  makeIcon,
   shadeElement,
 } from "./html-utils";
 
+type SortType = "number" | "text" | "date";
+
 export interface ColumnDefinition<ParentType, ChildType> {
   title: string;
+  sortType?: SortType;
   child?: boolean;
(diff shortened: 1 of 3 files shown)


ts/util/table-builder.ts
```diff
@@ -60,6 +76,71 @@ export class TableBuilder<ParentType, ChildType> {
...
+    descending: boolean,
+    dropdown: HTMLSelectElement
+  ) {
+    const option = document.createElement("option");
+    option.value = column.title + (descending ? "/desc" : "/asc");
+    option.setAttribute("data-column", column.title);
+    option.setAttribute("data-order", descending ? "desc" : "asc");
+    option.text =
+      column.title +
+      " - " +
```
Arcslogger, line 116: change to a semicolon? Seems to be the convention in other sorting menus

djcooke: I've seen colons, semicolons, dashes, no separator... There doesn't seem to be any established convention. I tried both ways and think that the dash looks better as it adds some extra space between the column name and ordering

Arcslogger (approved):

PR #31, opened 2022-07-11
[GLT-3722] Added dropdown menu controls

Arcslogger:


ts/util/html-utils.ts
```diff
@@ -1,5 +1,5 @@
 // adds a header cell to a table header
-export function addColumnHeader(
+export function makeColumnHeader(
```
djcooke, line 2: Please rename back to "add" for clarity since it does in fact add the element to the DOM

ts/util/html-utils.ts
```diff
@@ -22,7 +22,7 @@ export function addCell(tr: HTMLTableRowElement, index: number) {
 }
 
 // adds a styled link to an element
-export function addLink(container: HTMLElement, text: string, url: string) {
+export function makeLink(container: HTMLElement, text: string, url: string) {
```
djcooke, line 25: rename back to "add"

ts/util/html-utils.ts
```diff
@@ -71,3 +71,24 @@ export function makeIcon(name: string) {
...
+    "absolute hidden mt-1 w-fit rounded-md p-1 bg-grey-100 font-inter font-medium text-black text-12 drop-shadow-lg cursor-pointer";
+  return dropdownMenuContainer;
+}
+
+export function makeDropdownClickout() {
+  const clickout = document.createElement("button");
+  clickout.className =
+    "bg-transparent fixed inset-0 w-full h-full cursor-default hidden";
+  return clickout;
+}
```
djcooke, line 94: This is all dropdown-specific, so should be in `dropdown.ts`, and not exported

ts/util/table-builder.ts
```diff
@@ -77,52 +96,69 @@ export class TableBuilder<ParentType, ChildType> {
...
+      dropdownMenu.classList.toggle("hidden");
+      dropdownClickout.classList.toggle("hidden");
+    };
+    sortButton.onclick = toggleMenu;
+    dropdownClickout.onclick = toggleMenu;
+
+    dropdownContainer.appendChild(sortButton);
+    dropdownContainer.appendChild(dropdownClickout);
+    dropdownContainer.appendChild(dropdownMenu);
+    sortContainer.appendChild(dropdownContainer);
```
djcooke, line 147: This should all be handled by `makeDropdownMenu` in `dropdown.ts`. `makeDropdownMenu` should return the outermost element (`dropdownContainer`) that can be added to the DOM (`sortContainer`). All that should be in the table builder is a call to `makeDropdownMenu` providing the `DropdownOption`s.

ts/util/table-builder.ts
```diff
@@ -141,6 +177,12 @@ export class TableBuilder<ParentType, ChildType> {
     }
   }
 
+  private applySort(title: string, decending: boolean) {
+    this.sortColumn = title;
+    this.sortDescending = decending;
+    this.reload();
+  }
```
djcooke, line 184: Can collapse this into `addSortOption` since it's small and only used there

ts/util/dropdown.ts
```diff
@@ -0,0 +1,72 @@
...
+  const dropdownContainer = document.createElement("div");
+  const dropdownClickout = makeDropdownClickout();
+
+  const dropdownMenuContainer = makeDropdownMenuContainer();
+  DropdownOptions.forEach((option) => {
+    const li = document.createElement("li");
+    option.render(li);
+    li.addEventListener(
+      "click",
+      () => (dropdownButton.innerHTML = option.text)
```
djcooke, line 35: Should also close the menu here. It's currently staying open when you select an option

djcooke, line 35: This should close the dropdown as well. It's currently staying open when to select an option

djcooke (approved):

PR #36, opened 2022-07-18
[GLT-3711] Added dropdown filter controls

Arcslogger:
text-field filtering controls will be implemented in a separate PR

### DIFF
--- a/ts/util/table-builder.ts
+++ b/ts/util/table-builder.ts
@@ -60,7 +60,7 @@ class AcceptedFilter {
     key: string,
     value: string,
     container: HTMLElement,
-    reload: () => {}
+    onRemove: () => {}
   ) {
     this.key = key;
     this.value = value;
@@ -79,7 +79,7 @@ class AcceptedFilter {
     destroyFilterIcon.onclick = () => {
       this.valid = false;
       this.element.remove();
-      reload();
+      onRemove();
     };
     this.element.appendChild(destroyFilterIcon);
 
@@ -234,7 +234,7 @@ export class TableBuilder<ParentType, ChildType> {
       switch (filter.type) {
         case "dropdown":
           // create all submenu options
-          const onRemove = () => this.reload();
+          const reload = () => this.reload();
           const filterSuboptions = filter.values.map(
             (value) =>
               new BasicDropdownOption(value, (dropdown: Dropdown) => {
@@ -244,7 +244,7 @@ export class TableBuilder<ParentType, ChildType> {
                   filter.key,
                   value,
                   filterContainer,
-                  onRemove
+                  reload
                 );
                 // add filter to the menu bar
                 filterContainer.insertBefore(
@@ -252,7 +252,7 @@ export class TableBuilder<ParentType, ChildType> {
                   addFilterDropdown.getTag()
                 );
                 this.acceptedFilters.push(filterLabel);
-                onRemove();
+                reload();
               })
           );
           // add filter (& its submenu) to the parent filter menu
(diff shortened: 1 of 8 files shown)


ts/util/table-builder.ts
```diff
@@ -255,9 +373,16 @@ export class TableBuilder<ParentType, ChildType> {
...
     this.showLoading();
+    this.acceptedFilters = this.acceptedFilters.filter(
+      (filter) => filter.valid
+    );
     const response = await Rest.post(Rest.cases.query, {
       pageSize: this.pageSize,
       pageNumber: this.pageNumber,
+      filters: this.acceptedFilters.map((filter) => {
+        const mappedFilters = { key: filter.key, value: filter.value };
+        return mappedFilters;
```
djcooke, line 384: Could shorten to `return { key: filter.key, value: filter.value };`

ts/util/table-builder.ts
```diff
@@ -72,7 +124,7 @@ export class TableBuilder<ParentType, ChildType> {
 
   public build() {
     const topControlsContainer = document.createElement("div");
-    topControlsContainer.className = "flex mt-4 items-center";
+    topControlsContainer.className = "flex mt-4 items-center space-x-2";
```
djcooke, line 127: Change `items-center` to `items-top` to improve display when there are enough filters to cause wrapping

ts/util/table-builder.ts
```diff
@@ -169,10 +224,73 @@ export class TableBuilder<ParentType, ChildType> {
 
   private addFilterControls(container: HTMLElement) {
     const filterContainer = document.createElement("div");
-    filterContainer.classList.add("flex-auto");
-    container.appendChild(filterContainer);
+    filterContainer.classList.add("flex-auto", "space-x-2", "items-center");
```
djcooke, line 227: Would a `space-y-2` or something improve display when wrapping happens? Currently, the +filter button can overlap with the accepted filter elements

ts/util/table-builder.ts
```diff
@@ -33,13 +39,58 @@ export interface SortDefinition {
...
+    key: string,
+    value: string,
+    container: HTMLElement,
+    reload: () => {}
+  ) {
+    this.key = key;
+    this.value = value;
+    this.element = document.createElement("span");
+    this.element.className =
+      "font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md cursor-default";
```
djcooke, line 74: add `inline-block` to keep text together when wrapping filters

ts/util/table-builder.ts
Arcslogger: image
space-y-2 messes up the top margins unfortunately

djcooke: Changing the accepted filter elements to inline-block actually fixes the spacing too

ts/util/table-builder.ts
```diff
@@ -239,31 +234,31 @@ export class TableBuilder<ParentType, ChildType> {
       switch (filter.type) {
         case "dropdown":
           // create all submenu options
-          const reload = () => this.reload();
+          const onRemove = () => this.reload();
```
djcooke: Renamed in the wrong place. Think of the different contexts:

* In `AcceptedFitler`, the function is just a callback. `AcceptedFilter` doesn't know what the function will do, so the name here should be "onRemove"
* Here, we know what the function is doing - reloading the table. It's being used when adding filters as well as removing, so "onRemove" is a misleading name here and "reload" makes sense

djcooke (approved):
Don't forget to create a ticket for the dropdown open/close icons

mlaszloffy (approved):

PR #37, opened 2022-07-20
[GLT-3731] added text input filtering controls

Arcslogger:


mlaszloffy (approved):

ts/util/table-builder.ts
```diff
@@ -255,6 +256,11 @@ export class TableBuilder<ParentType, ChildType> {
                 reload();
               })
           );
+          if (!filterSuboptions) {
+            throw new Error(
+              `Dropdown filter ${filter.key} has no menu options`
+            );
+          }
```
djcooke, line 263: Move this check above the `const filterSuboptions...` line and change to `if (!filter.values || !filter.values.length)`. Then you don't need the `?` in `filter.values.map`, and it'll also protect against `filter.values` being an empty array

ts/util/text-input.ts
```diff
@@ -0,0 +1,68 @@
...
+        textInputClickout.classList.toggle("hidden");
+        onClose(this);
+      } else {
+        window.setTimeout(() => this.textField.focus(), 0);
+      }
+    };
+    submitIcon.onclick = submitTextInput;
+    textInputClickout.onclick = submitTextInput;
+    this.textField.addEventListener("keypress", (event) => {
+      console.log(event.key);
```
djcooke, line 37: Remove

ts/util/text-input.ts
```diff
@@ -0,0 +1,68 @@
...
+      "relative",
+      "z-10"
+    );
+
+    const submitTextInput = () => {
+      if (this.textField.value) {
+        textInputClickout.classList.toggle("hidden");
+        onClose(this);
+      } else {
+        window.setTimeout(() => this.textField.focus(), 0);
```
djcooke, line 31: I don't think `setTimeout` is required here?

ts/util/table-builder.ts
```diff
@@ -271,7 +277,31 @@ export class TableBuilder<ParentType, ChildType> {
           );
           break;
         case "text":
```
djcooke, line 279: `addFilterControls` has gotten pretty long. Maybe separate the cases out into their own functions? e.g.

```
function addFilterControls(...) {
  ...
  switch (filterType) {
    case "dropdown":
      addDropdownFilter(...);
      return;
    case "text":
      addTextFilter(...);
      return;
    default:
      ...
  }
}
```

djcooke (approved):

PR #44, opened 2022-07-25
reworked filter bar spacing and added paging icon

Arcslogger:


ts/util/table-builder.ts
```diff
@@ -286,48 +274,59 @@ export class TableBuilder<ParentType, ChildType> {
             filter.title,
             filter.key,
             value,
-            labelContainer,
             reload
           );
-          // add filter to the menu bar
-          labelContainer.appendChild(filterLabel.element);
+          console.log(filterContainer.lastChild);
```
djcooke, line 279: remove

ts/util/table-builder.ts
```diff
@@ -286,48 +274,59 @@ export class TableBuilder<ParentType, ChildType> {
...
   ) {
     const onClose = (textInput: TextInput) => {
       const filterLabel = new AcceptedFilter(
         filter.title,
         filter.key,
         textInput.getValue(),
-        labelContainer,
         reload
       );
+      console.log(filterContainer.lastChild);
```
djcooke, line 315: remove

src/main/resources/static/css/output.css
```diff
@@ -670,6 +670,19 @@ Ensure the default browser behavior of the `hidden` attribute.
...
+}
+
+.gap-x-2{
+  -moz-column-gap: 0.5rem;
+       column-gap: 0.5rem;
+}
+
+.gap-y-2{
+  row-gap: 0.5rem;
+}
```
djcooke, line 684: It looks like these aren't being used, so probably need to regenerate?

mlaszloffy (approved):

djcooke (approved):

PR #49, opened 2022-07-27
[GLT-3737] added icon legend

Arcslogger:


Arcslogger:
- Parity w/figma: legend toggle button now shows up at the bottom of each table
- Fixed popup disappearing when re-opened on detail pages

ts/util/legend.ts
```diff
@@ -0,0 +1,159 @@
...
+import { makeIcon } from "./html-utils";
+
+export class Legend {
+  private container: HTMLElement;
+  constructor() {
+    // outer container
+    this.container = document.createElement("div");
+    this.container.className =
+      "text-black block flex flex-col bg-white rounded-xl fixed border-2 border-grey-200 z-40 drop-shadow-lg";
+    this.container.id = "legend-container";
```
djcooke, line 11: Make this a constant since it's used in at least 2 places. Will help to avoid bugs if things get renamed/moved

ts/util/legend.ts
```diff
@@ -0,0 +1,159 @@
...
+
+      // disable window moving if the cursor is out of window
+      if (
+        e.clientX < 0 ||
+        e.clientX > window.innerWidth ||
+        e.clientY < 0 ||
+        e.clientY > window.innerHeight
+      ) {
+        return;
+      }
```
djcooke, line 100: Remove this and it'll have the behaviour you wanted - it'll be bound to the window, but will still move on any axis that you're in bounds of

ts/util/legend.ts
```diff
@@ -0,0 +1,159 @@
...
+      previousPosX = e.clientX;
+      previousPosY = e.clientY;
+      // When the mouse is let go, call the closing event
+      document.onmouseup = closeDragElement;
+      // call a function whenever the cursor moves
+      document.onmousemove = elementDrag;
+    };
+
+    const elementDrag = (e: MouseEvent) => {
+      // Prevent any default action on this element (you can remove if you need this element to perform its default action)
```
djcooke, line 89: Seems like this was copied from somewhere? Remove the comment that's not relevant here, but add a reference the original source

ts/util/legend.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    // centering window, setTimeout is required for offsetHeight and offsetWidth to load
+    // TODO: fix flash introduced by timeout
+    window.setTimeout(() => {
+      this.container.style.top =
+        (window.innerHeight - this.container.offsetHeight) / 2 + "px";
+      this.container.style.left =
+        (window.innerWidth - this.container.offsetWidth) / 2 + "px";
+    }, 0);
```
djcooke, line 61: Can this be done by making it hidden when adding to the DOM, then make it visible here after positioning it?

ts/util/table-builder.ts
```diff
@@ -374,6 +381,23 @@ export class TableBuilder<ParentType, ChildType> {
...
+    legendButton.onclick = () => {
+      const legendWindow = document.getElementById("legend-container");
+      if (!legendWindow) {
+        const legendContainer = new Legend();
+        this.container.appendChild(legendContainer.getTag());
+      } else {
+        legendWindow.remove();
+      }
+    };
+    container.appendChild(legendButton);
```
djcooke, line 398: Instead of having this code here, could there be a `toggleLegend` function in `legend.ts`? Don't even need to pass it the container since you can append it to the document body instead and it shouldn't make any difference

ts/util/legend.ts
Arcslogger: I can't tell, but it seems to be less frequent. Issue still persists though

ts/util/legend.ts
Arcslogger: what did you mean by this? Did you want me to cite the code I used?

ts/util/legend.ts
Arcslogger: I prefer just disabling moving the window entirely when the cursor is OOB for now. With this block removed the window will move on both axes, including the one OOB.

Ideally, it only moves in the axis thats in bounds. (e.g if my cursor is OOB on the left, then the window still moves on the y-axis but stays locked on the x-axis until the cursor goes back into the window). [This demo](https://www.columbia.edu/~njn2118/journal/2019/4/26.htmll) gets pretty close to what I'm looking for, but they seem to use some slightly different math that I couldn't crack before I made the PR :(

djcooke: The clamp functions prevent the window from going out of bounds, so it will only move on an axis that is in bounds if this is removed. I have tried and verified this

ts/util/legend.ts
djcooke: Yes, include a link in comment when copying a significant chunk of code

ts/util/legend.ts
djcooke: I just tried and it does solve the problem. Using "hidden" creates a different problem - the container has no height or width if it's hidden, so this doesn't calculate the position correctly. You can use "invisible" instead and both problems are solved - add "invisible" to the container when first created, and remove it after setting the position here.

mlaszloffy (approved):

ts/util/table-builder.ts
```diff
@@ -148,14 +148,18 @@ export class TableBuilder<ParentType, ChildType> {
     const bottomControlsContainer = document.createElement("div");
     bottomControlsContainer.className =
       "flex flex-row-reverse mt-4 items-top space-x-2";
-    this.addLegendControls(bottomControlsContainer);
+    this.addActionButtons(bottomControlsContainer);
     this.container.appendChild(bottomControlsContainer);
 
     this.load();
     // TODO: add action buttons
```
djcooke, line 155: move/remove comment since you have implemented the part that goes here

ts/util/legend.ts
```diff
@@ -1,14 +1,15 @@
 import { qcStatuses } from "../data/qc-status";
 import { makeIcon } from "./html-utils";
 
+export const legendID = "legend-container";
```
djcooke, line 4: * Should be `legendId` for consistency with other ID variable names throughout Dimsum
* No need to export

ts/util/legend.ts
```diff
@@ -1,14 +1,15 @@
...
 export class Legend {
```
djcooke, line 5: No longer needs to be exported

djcooke (approved):

PR #55, opened 2022-08-04
[GLT-3739] added autocomplete to text input filters

Arcslogger:
also removed case sensitivity in the text input filters

### DIFF
--- a/src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
@@ -117,8 +117,34 @@ public class CaseLoader {
         refreshTimer.record(System.currentTimeMillis() - startTimeMillis, TimeUnit.MILLISECONDS);
       }
       log.debug(String.format("Completed loading %d cases.", cases.size()));
-      return new CaseData(cases, runsByName, afterTimestamp);
+      return new CaseData(cases, runsByName, afterTimestamp, getAssayNames(cases),
+          getRequisitionNames(requisitionsById), getProjectNames(projectsByName),
+          getDonorNames(donorsById));
     }
+  }
+
+  private Set<String> getAssayNames(List<Case> cases) {
+    Set<String> assayNames =
+        new HashSet<String>(cases.stream().map(Case::getAssayName).collect(Collectors.toSet()));
+    return assayNames;
+  }
+
+  private Set<String> getRequisitionNames(Map<Long, Requisition> requisitionsById) {
+    Set<String> sampleNames = new HashSet<String>(
+        requisitionsById.values().stream().map(Requisition::getName).collect(Collectors.toSet()));
+    return sampleNames;
+  }
+
+  private Set<String> getProjectNames(Map<String, Project> projectsByName) {
+    Set<String> projectNames =
+        new HashSet<String>(projectsByName.keySet().stream().collect(Collectors.toSet()));
+    return projectNames;
+  }
+
+  private Set<String> getDonorNames(Map<String, Donor> donorsById) {
+    Set<String> donorNames = new HashSet<String>(
+        donorsById.values().stream().map(Donor::getName).collect(Collectors.toSet()));
+    return donorNames;
   }
 
   protected FileReader getProjectReader() throws FileNotFoundException {
--- a/src/main/java/ca/on/oicr/gsi/dimsum/data/CaseData.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/data/CaseData.java
@@ -7,6 +7,7 @@ import java.time.ZonedDateTime;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.annotation.concurrent.Immutable;
 
 @Immutable
@@ -15,12 +16,23 @@ public class CaseData {
   private final List<Case> cases;
   private final Map<String, RunAndLibraries> runsByName;
   private final ZonedDateTime timestamp;
+  private final Set<String> assayNames;
+  private final Set<String> requisitionNames;
+  private final Set<String> projectsNames;
+  private final Set<String> donorNames;
+
+
 
   public CaseData(List<Case> cases, Map<String, RunAndLibraries> runsByName,
-      ZonedDateTime timestamp) {
+      ZonedDateTime timestamp, Set<String> assays, Set<String> requisitions,
+      Set<String> projects, Set<String> donors) {
     this.cases = unmodifiableList(cases);
     this.runsByName = Collections.unmodifiableMap(runsByName);
     this.timestamp = requireNonNull(timestamp);
+    this.assayNames = Collections.unmodifiableSet(assays);
+    this.requisitionNames = Collections.unmodifiableSet(requisitions);
+    this.projectsNames = Collections.unmodifiableSet(projects);
+    this.donorNames = Collections.unmodifiableSet(donors);
   }
 
   public List<Case> getCases() {
@@ -33,6 +45,22 @@ public class CaseData {
 
   public ZonedDateTime getTimestamp() {
     return timestamp;
+  }
+
+  public Set<String> getAssayNames() {
+    return assayNames;
+  }
+
+  public Set<String> getRequisitionNames() {
+    return requisitionNames;
+  }
+
+  public Set<String> getProjectNames() {
+    return projectsNames;
+  }
+
+  public Set<String> getDonorNames() {
+    return donorNames;
   }
 
 }
--- a/src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
@@ -111,23 +111,26 @@ public class CaseService {
   }
 
   public Set<String> getMatchingAssayNames(String prefix) {
-    return getMatchingNames(caseData.getAssayNames(), prefix);
+    return caseData.getAssayNames().stream()
+        .filter(s -> s.toLowerCase().startsWith(prefix.toLowerCase()))
+        .collect(Collectors.toSet());
   }
 
   public Set<String> getMatchingRequisitionNames(String prefix) {
-    return getMatchingNames(caseData.getRequisitionNames(), prefix);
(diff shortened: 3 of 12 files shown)


blueprint.md
```diff
@@ -0,0 +1,53 @@
+# Notes from discussion meeting, to be deleted on final PR
```
djcooke: You said it 😆

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -117,10 +117,36 @@ public CaseData load(ZonedDateTime previousTimestamp) throws DataParseException,
...
     }
   }
 
+  private Set<String> getAssayNames(List<Case> cases) {
+    Set<String> assayNames =
+        new HashSet<String>(cases.stream().map(Case::getAssayName).collect(Collectors.toSet()));
+    return assayNames;
+  }
+
+  private Set<String> getRequsitionNames(Map<Long, Requisition> requisitionsById) {
```
djcooke, line 132: Typo in "Requisition"

src/main/java/ca/on/oicr/gsi/dimsum/data/CaseData.java
```diff
@@ -15,12 +16,23 @@ public class CaseData {
...
+  private final Set<String> RequisitionNames;
+  private final Set<String> ProjectsNames;
+  private final Set<String> DonorNames;
+
+
 
   public CaseData(List<Case> cases, Map<String, RunAndLibraries> runsByName,
-      ZonedDateTime timestamp) {
+      ZonedDateTime timestamp, Set<String> Assays, Set<String> Requisitions,
+      Set<String> Projects, Set<String> Donors) {
```
djcooke, line 28: field and parameter names should start with lowercase

src/main/java/ca/on/oicr/gsi/dimsum/data/CaseData.java
```diff
@@ -15,12 +16,23 @@ public class CaseData {
...
     this.cases = unmodifiableList(cases);
     this.runsByName = Collections.unmodifiableMap(runsByName);
     this.timestamp = requireNonNull(timestamp);
+    this.AssayNames = Assays;
+    this.RequisitionNames = Requisitions;
+    this.ProjectsNames = Projects;
+    this.DonorNames = Donors;
```
djcooke, line 35: Use `Collections.unmodifiableSet` to make these immutable (similar to `unmodifiableMap` used above)

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -111,6 +110,30 @@ public TableData<Case> getCases(int pageSize, int pageNumber, CaseSort sort, boo
...
+    return caseData.getProjectNames().stream()
+        .filter(s -> s.toLowerCase().startsWith(prefix.toLowerCase()))
+        .collect(Collectors.toSet());
+  }
+
+  public Set<String> getMatchingDonorNames(String prefix) {
+    return caseData.getDonorNames().stream()
+        .filter(s -> s.toLowerCase().startsWith(prefix.toLowerCase()))
+        .collect(Collectors.toSet());
+  }
```
djcooke, line 135: Add a private `getMatchingNames(Set<String>, String)` method to reduce the repetition here

ts/util/requests.ts
```diff
@@ -15,6 +15,16 @@ export function post(url: string, body: any) {
   });
 }
 
+export function get(base: string, body: string) {
```
djcooke, line 18: rename "base" -> "url" since this is the full URL. "Base URL" usually means just the server name

ts/util/table-builder.ts
```diff
@@ -42,6 +42,7 @@ export interface FilterDefinition {
   key: string; // internal use
   type: FilterType; // either text or dropdown
   values?: string[]; // required for dropdown filters
+  url?: string; // required for text filters w/autocomplete
```
djcooke, line 45: rename "autocompleteUrl" for clarity

ts/util/requests.ts
```diff
@@ -15,6 +15,16 @@ export function post(url: string, body: any) {
...
+  let headers: any = {
+    "Content-Type": "application/json",
+  };
+  return fetch(`${base}?q=${encodeURIComponent(body)}`, {
```
djcooke, line 22: `get` should be able to handle any parameters - not just `q`. There should be a `parameters: Record<string, string>` parameter, so this:

```
get("https://dimsum.com/rest/test", {
  "q": "some text",
  "otherParam": 1234
});
```

results in: `GET https://dimsum.com/rest/test?q=some%20text&otherParam=1234`

You can use `new URLSearchParams(parmeters).toString()` to generate everything after the `?`

ts/util/text-input.ts
```diff
@@ -58,9 +72,31 @@ export class TextInput {
     }
   }
 
-  public getTag() {
+  private async loadAutocomplete(queryUrl: string) {
+    const val = this.textField.value;
+    if (val.length < 1) {
+      // results are only displayed for input > **2** characters
+      return;
```
djcooke, line 79: condition and comment don't seem to match?

ts/util/table-builder.ts
```diff
@@ -331,9 +333,18 @@ export class TableBuilder<ParentType, ChildType> {
...
+        );
+      }
+      const filterTextInput = new TextInput(filter.title, onClose, filter.url);
+      filterContainer.insertBefore(
+        filterTextInput.getContainerTag(),
+        filterContainer.lastChild
+      );
       filterContainer.insertBefore(
-        filterTextInput.getTag(),
+        filterTextInput.getDatalistTag(),
```
djcooke, line 347: The datalist should be inside the container, so it doesn't need to be added/removed separately

ts/util/text-input.ts
```diff
@@ -15,7 +25,9 @@ export class TextInput {
     this.container.className =
       "font-inter font-medium text-12 text-black px-2 py-1 rounded-md ring-2 ring-offset-1 ring-red inline-block flex-auto items-center space-x-2";
     this.textField.className =
-      "ring-0 border-0 outline-0 rounded-sm relative z-10 px-1 w-4";
+      "ring-0 border-0 outline-0 rounded-sm relative z-10 px-1";
+    this.textField.style.minWidth = "20ch";
+    this.textField.style.width = "20ch";
```
djcooke, line 30: Shouldn't be using inline styles. Use Tailwind class to set minWidth and don't set width

Arcslogger: the snipped used to increase width dynamically doesn't play well with tailwind's min-w unfortunately :( Using .style is messier but it's the only way I've managed to get it to work

ts/util/text-input.ts
Arcslogger: doc here is correct due to the nature of how keypress events work (i moved autocomplete querys to the keypress event listener so it doesnt autocomplete again once you select from the datalist)

ts/util/text-input.ts
djcooke: I played around with it and found a solution that works without inline styles:

Here (in the constructor):

* add class `min-w-[150px]`
* set size: `this.textField.setAttribute("size", "10");`

In `keydown`:

* `this.textField.setAttribute("size", Math.max(this.textField.value.length, 10).toString());`

ts/util/text-input.ts
djcooke: Oh, I see. Can improve this by using the `input` event instead of `keydown` and moving the `loadAutocomplete` call there. Then you'll get the actual full value here, so you can check `val.length < 2`

Should also clear the datalist before returning here. Otherwise, if you type "WG", clear the input and then enter "T" it still shows the entries for "WG"

Arcslogger: moving `loadAutocomplete` to a `input` event reintroduces the duped autocomplete behaviour though. I had that originally before moving it to the `keydown` event

ts/util/requests.ts
```diff
@@ -15,11 +15,11 @@ export function post(url: string, body: any) {
   });
 }
 
-export function get(base: string, body: string) {
+export function get(url: string, params: any) {
```
djcooke, line 18: `params` should have type `Record<string, string>`. This forces the keys and values to all be strings. Most other values that would be accepted as `any` would result in errors

ts/util/table-builder.ts
```diff
@@ -324,7 +324,7 @@ export class TableBuilder<ParentType, ChildType> {
         reload
       );
       textInput.getContainerTag().remove();
-      textInput.getDatalistTag().remove();
+      //textInput.getDatalistTag().remove();
```
djcooke, line 327: remove commented-out code

ts/util/table-builder.ts
```diff
@@ -333,20 +333,24 @@ export class TableBuilder<ParentType, ChildType> {
...
       );
       filterContainer.insertBefore(
-        filterTextInput.getDatalistTag(),
+        filterTextInput.getContainerTag(),
         filterContainer.lastChild
       );
+      // filterContainer.insertBefore(
+      //   filterTextInput.getDatalistTag(),
+      //   filterContainer.lastChild
+      // );
```
djcooke, line 353: remove commented-out code

ts/util/text-input.ts
```diff
@@ -74,15 +77,15 @@ export class TextInput {
 
   private async loadAutocomplete(queryUrl: string) {
     const val = this.textField.value;
+    this.datalist.replaceChildren(); // empties autocompelte
```
djcooke, line 80: typo in comment

ts/util/text-input.ts
```diff
@@ -58,9 +82,30 @@ export class TextInput {
     }
   }
 
-  public getTag() {
+  private async loadAutocomplete(queryUrl: string) {
+    const val = this.textField.value;
+    this.datalist.replaceChildren(); // empties autocomplete
+    if (val.length < 2) {
+      // results are only displayed for input > **2** characters
```
djcooke, line 89: Can change to work after 1 character. It's really fast, so not a problem. We can adjust in the future if necessary

djcooke (approved):

mlaszloffy (approved):

PR #57, opened 2022-08-09
[GLT-3745] added more detailed tooltips

Arcslogger:


ts/data/case.ts
```diff
@@ -532,16 +539,73 @@ function addSampleIcons(samples: Sample[], fragment: DocumentFragment) {
...
+  const projectLink = makeNameDiv(
+    sample.project,
+    urls.miso.project(sample.project),
+    urls.dimsum.project(sample.project)
+  );
+  bottomContainer.appendChild(projectLabel);
+  // donor links
+  const donorLabel = document.createElement("span");
+  donorLabel.innerHTML = "Donor:";
+  const donorLinks = document.createElement("div");
```
djcooke, line 583: Rename for clarity - contains more than just links (external name will never be a link)

ts/data/case.ts
```diff
@@ -532,16 +539,73 @@ function addSampleIcons(samples: Sample[], fragment: DocumentFragment) {
...
+  );
+  // project links
+  const projectLabel = document.createElement("span");
+  projectLabel.innerHTML = "Project:";
+  const projectLink = makeNameDiv(
+    sample.project,
+    urls.miso.project(sample.project),
+    urls.dimsum.project(sample.project)
+  );
+  bottomContainer.appendChild(projectLabel);
```
djcooke, line 579: abstract to an `addRow(container, label, name, misoUrl, dimsumUrl)` function that can be reused for the donor name/links below and requisition later on. Add both label and name div to the container in this function

ts/data/case.ts
```diff
@@ -532,16 +539,73 @@ function addSampleIcons(samples: Sample[], fragment: DocumentFragment) {
...
+  donorLinks.appendChild(externalDonorName);
+  // requisition links TODO
+  const requisitionLabel = document.createElement("span");
+  requisitionLabel.innerHTML = "Requisition:";
+
+  bottomContainer.appendChild(projectLabel);
+  bottomContainer.appendChild(projectLink);
+  bottomContainer.appendChild(donorLabel);
+  bottomContainer.appendChild(donorLinks);
+  bottomContainer.appendChild(requisitionLabel);
```
djcooke, line 603: remove additions that already happen elsewhere, move others up to where the element is created

ts/data/case.ts
```diff
@@ -566,10 +630,22 @@ function addRequisitionIcon(
   fragment: DocumentFragment
 ) {
   const icon = makeIcon(status.icon);
-  icon.title = `${requisition.name}: ${status.label}`; // TODO: more detailed popup
+  addToolTip(icon, makeRequisitionTooltip(requisition));
   fragment.appendChild(icon);
 }
 
+function makeRequisitionTooltip(requisition: Requisition) {
+  const toolTipConatiner = new DocumentFragment();
```
djcooke, line 638: typo

ts/util/html-utils.ts
```diff
@@ -102,3 +102,12 @@ export function makeNameDiv(name: string, misoUrl: string, dimsumUrl?: string) {
...
   return div;
 }
+
+export function addToolTip(target: HTMLElement, contents: DocumentFragment) {
+  target.classList.add("group");
+  const tipContainer = document.createElement("div");
+  tipContainer.className =
+    "absolute bg-white w-fit h-fit p-2 border-2 border-dotted border-green-200 font-inter font-14 font-medium group-hover:visible invisible";
+  tipContainer.appendChild(contents);
+  target.appendChild(tipContainer);
```
djcooke, line 112: This is effectively adding a `div` to an `i`, which is against the HTML5 spec - `i` should only contain phrasing content. You probably want to wrap the `i` in a `div` and add `tipContainer` to that div too

djcooke (commented):
* If you scroll horizontally, the tooltips no longer line up properly. You can fix this by making the parent element `position: relative`, though that causes other problems
* The original design on Figma had the tooltips centred beneath the icon. This has them hanging to the right. I think I prefer hanging to the right, but it would be nice to have them pushed a little to the left, so there's more of an overlap

ts/data/case.ts
```diff
@@ -629,21 +619,45 @@ function addRequisitionIcon(
...
+  const tooltipContainer = new DocumentFragment();
+  tooltipContainer.appendChild(
     makeNameDiv(
       requisition.name,
       urls.miso.requisition(requisition.id),
       urls.dimsum.requisition(requisition.id)
     )
   );
-  return toolTipConatiner;
+  return tooltipContainer;
```
djcooke, line 638: If you change the type of `contents` in `makeTooltip` to `Node`, it'll accept an `HTMLElement` or `DocumentFragment`, so you can just return `makeNameDiv(...)` here

ts/util/table-builder.ts
```diff
@@ -126,7 +126,8 @@ export class TableBuilder<ParentType, ChildType> {
     this.container.appendChild(topControlsContainer);
 
     const tableContainer = document.createElement("div");
-    tableContainer.className = "mt-4 overflow-auto";
+    tableContainer.className =
+      "mt-4 overflow-x-auto overflow-y:hidden relative";
```
djcooke, line 130: `overflow-y:hidden` should  be `overflow-y-hidden`

ts/util/html-utils.ts
```diff
@@ -103,15 +103,15 @@ export function makeNameDiv(name: string, misoUrl: string, dimsumUrl?: string) {
   return div;
 }
 
-export function makeTooltip(target: HTMLElement, contents: DocumentFragment) {
+export function makeTooltip(target: HTMLElement, contents: Node) {
   const tooltipContainer = document.createElement("div");
-  tooltipContainer.className = "inline";
-  target.classList.add("group", "hover:text-green-200");
+  tooltipContainer.className = "inline group";
+  target.classList.add("hover:text-green-200");
```
djcooke, line 109: Highlight is lost when you mouse into the tooltip. Change to `group-hover`?

ts/util/table-builder.ts
```diff
@@ -125,9 +125,10 @@ export class TableBuilder<ParentType, ChildType> {
     this.addPagingControls(topControlsContainer);
     this.container.appendChild(topControlsContainer);
 
+    const tableWrapper = document.createElement("div");
+    tableWrapper.className = "relative";
```
djcooke, line 129: Please add a comment to explain why this is necessary (separating `relative` and `overflow:hidden` to allow the tooltips to overflow without additional scrollbars)

ts/util/table-builder.ts
```diff
@@ -125,9 +125,10 @@ export class TableBuilder<ParentType, ChildType> {
...
     const tableContainer = document.createElement("div");
-    tableContainer.className =
-      "mt-4 overflow-x-auto overflow-y-hidden relative";
+    tableContainer.className = "mt-4 overflow-x-auto overflow-x-hidden";
```
djcooke, line 131: Accidentally changed `overflow-y-hidden` to `overflow-x-hidden`? Entire page scrolls horizontally to table width now, but the table is also cut off so there's nothing visible in that additional space

Arcslogger:
not sure how we're getting a `nullPointer` on the tests rn...?

ts/util/table-builder.ts
```diff
@@ -124,11 +124,11 @@ export class TableBuilder<ParentType, ChildType> {
     this.addFilterControls(topControlsContainer);
     this.addPagingControls(topControlsContainer);
     this.container.appendChild(topControlsContainer);
-
+    // tableWrapper allows tooltips to overflow without using additional scrollbars
     const tableWrapper = document.createElement("div");
-    tableWrapper.className = "relative";
+    // tableWrapper.className = "relative";
```
djcooke, line 129: `tableWrapper` is no longer needed

ts/data/tooltip.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.tooltipContainer = this.getTootipContainer();
+
+    target.onmouseenter = () => {
+      this.mouseOnIcon = true;
+      const targetY = this.getOffset(target).bottom;
+      const targetX = this.getOffset(target).left;
+      // empties tooltip container
+      this.tooltipContainer.replaceChildren();
+      // node needs to be cloned since appendChild removes it
+      this.tooltipContainer.appendChild(contents.cloneNode(true));
```
djcooke, line 18: Can add the contents in `replaceChildren` instead of using `appendChild` afterwards

ts/data/tooltip.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  constructor(target: HTMLElement, contents: Node) {
+    this.mouseOnIcon = false;
+    this.mouseOnToolTip = false;
+    this.tooltipContainer = this.getTootipContainer();
+
+    target.onmouseenter = () => {
+      this.mouseOnIcon = true;
+      const targetY = this.getOffset(target).bottom;
+      const targetX = this.getOffset(target).left;
```
djcooke, line 14: Calculating the same thing twice

ts/util/tooltip.ts
```diff
@@ -0,0 +1,97 @@
...
+  private getTootipContainer = () => {
+    let div = document.getElementById("tooltip");
+    if (div) {
+      return div;
+    }
+    div = document.createElement("div");
+    div.id = "tooltip";
+    div.classList.add(
+      "absolute",
+      "block",
```
djcooke, line 31: Unnecessary - divs are block by default

src/main/resources/static/css/output.css
```diff
@@ -510,6 +510,10 @@ Ensure the default browser behavior of the `hidden` attribute.
   visibility: hidden;
 }
 
+.static{
```
djcooke, line 513: Probably need to regenerate as I don't think this is used anywhere

ts/util/dropdown.ts
```diff
@@ -101,6 +101,6 @@ function makeDropdownButton() {
 function makeDropdownMenuContainer() {
   const dropdownMenuContainer = document.createElement("menu");
   dropdownMenuContainer.className =
-    "absolute hidden mt-2 w-fit rounded-md p-1 bg-grey-100 font-inter font-medium text-black text-12 drop-shadow-lg cursor-pointer";
+    "absolute hidden mt-2 w-fit rounded-md p-1 bg-grey-100 font-inter font-medium text-black text-12 drop-shadow-lg cursor-pointer z-10";
```
djcooke, line 104: Is the z-index still necessary?

ts/util/html-utils.ts
```diff
@@ -85,7 +85,7 @@ export function addIconButton(container: HTMLElement, iconName: string) {
 export function makeClickout() {
   const clickout = document.createElement("button");
   clickout.className =
-    "bg-transparent fixed inset-0 w-full h-full cursor-default hidden";
+    "bg-transparent fixed inset-0 w-full h-full cursor-default hidden z-10";
```
djcooke, line 88: Is the z-index still necessary?

ts/util/table-builder.ts
```diff
@@ -124,9 +124,8 @@ export class TableBuilder<ParentType, ChildType> {
     this.addFilterControls(topControlsContainer);
     this.addPagingControls(topControlsContainer);
     this.container.appendChild(topControlsContainer);
-
     const tableContainer = document.createElement("div");
-    tableContainer.className = "mt-4 overflow-auto";
+    tableContainer.className = "mt-4 overflow-x-auto overflow-y-hidden";
```
djcooke, line 128: Can revert this to `overflow-auto` to make it more obvious if we ever have another issue like the one with the tooltips clipping vs. creating vertical scrollbars

ts/util/tooltip.ts
```diff
@@ -0,0 +1,97 @@
...
+  private update() {
+    if (!this.activeTarget) {
+      throw new Error("Tooltip called with no target elements");
+    }
+    if (this.mouseOnIcon || this.mouseOnToolTip) {
+      this.tooltipContainer.classList.remove("invisible");
+      this.activeTarget.classList.add("text-green-200");
+    } else {
+      this.tooltipContainer.classList.add("invisible");
+      this.activeTarget.classList.remove("text-green-200");
```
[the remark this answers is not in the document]
Arcslogger: doing this seems to break the hover behaviour, mousing over the tooltip div doesn't work

src/main/resources/static/css/output.css
Arcslogger: regenerated & also tried manually deleting, but tailwind seems persistent on having this in the output file...

ts/util/tooltip.ts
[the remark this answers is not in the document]
djcooke: Ah, I guess because of mouseleave firing before mouseenter. Oh well. Thanks for trying 👍

src/main/resources/static/css/output.css
djcooke: 🤷‍♂️

djcooke (approved):

mlaszloffy (approved):

PR #66, opened 2022-09-06
[GLT-3757] Identify text fields

mengjenn:
Add tool tips for identifying non-intuitive text fields (group ids, external names, etc.). As per [GLT-3757](https://jira.oicr.on.ca/browse/GLT-3757).

### DIFF
--- a/ts/data/case.ts
+++ b/ts/data/case.ts
@@ -6,6 +6,7 @@ import {
   addMisoIcon,
   makeNameDiv,
   addNaText,
+  addTextDiv,
 } from "../util/html-utils";
 import { urls } from "../util/urls";
 import { siteConfig } from "../util/site-config";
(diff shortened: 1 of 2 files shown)


ts/data/case.ts
```diff
@@ -172,9 +172,9 @@ export const caseDefinition: TableDefinition<Case, Test> = {
         fragment.appendChild(nameDiv);
 
         const externalNameDiv = document.createElement("div");
-        externalNameDiv.appendChild(
-          document.createTextNode(kase.donor.externalName)
-        );
+        externalNameDiv.appendChild(document.createTextNode(kase.donor.externalName));
+        const tooltipInstance = Tooltip.getInstance();
+        tooltipInstance.addTarget(externalNameDiv, makeNameDiv("External Name", ""));
```
djcooke, line 177: Can use `document.createTextNode` instead of `makeNameDiv` since it doesn't need all of the links. Same in the other instances

mengjenn: Done

ts/data/case.ts
```diff
@@ -184,6 +189,10 @@ export const caseDefinition: TableDefinition<Case, Test> = {
               (kase.timepoint ? " " + kase.timepoint : "")
           )
         );
+        tooltipInstance.addTarget(
+          tumourDetailDiv,
+          document.createTextNode("Timepoint")
```
djcooke, line 194: Hmm... This line shows 2-3 different attributes, and timepoint isn't always included. Can you change this tooltip to look like below?

> Tissue Origin: XY
> Tissue Type: Z
> Timepoint: asdf

and omit the timepoint line if the sample has no timepoint

ts/data/case.ts
```diff
@@ -191,7 +191,11 @@ export const caseDefinition: TableDefinition<Case, Test> = {
         );
         tooltipInstance.addTarget(
           tumourDetailDiv,
-          document.createTextNode("Timepoint")
+          document.createTextNode(
+            `Tissue Origin: ${kase.tissueOrigin}\n` +
+              `Tissue Type: ${kase.tissueType}` +
+              (kase.timepoint ? `\nTimepoint: ${kase.timepoint}` : "")
```
djcooke, line 197: These don't render as line breaks in the browser. The easiest way to achieve this is to put each line in its own `div` as they are block by default and will stack vertically

ts/data/case.ts
```diff
@@ -184,6 +189,25 @@ export const caseDefinition: TableDefinition<Case, Test> = {
...
+          document.createTextNode(`Tissue Type: ${kase.tissueType}`)
+        );
+        tooltipDiv.appendChild(tissueTypeDiv);
+        const timepointDiv = document.createElement("div");
+        timepointDiv.appendChild(
+          document.createTextNode(
+            kase.timepoint ? `\nTimepoint: ${kase.timepoint}` : ""
+          )
+        );
+        tooltipDiv.appendChild(timepointDiv);
```
djcooke, line 209: Check whether the case has a timepoint and only create the timepoint div if so.

Can also create an `addTextDiv(text, container)` function that creates the div, sets the text, and adds it to the container (`tooltipDiv`) to reduce the repetition

mlaszloffy (approved):

PR #70, opened 2022-09-09
[GLT-3762] Show run-library lane number

mengjenn:
As per [GLT-3762](https://jira.oicr.on.ca/browse/GLT-3762)

### DIFF
--- a/ts/data/case.ts
+++ b/ts/data/case.ts
@@ -589,7 +589,9 @@ function makeSampleTooltip(sample: Sample) {
   if (sample.run) {
     topContainer.appendChild(
       makeNameDiv(
-        sample.run.name,
+        sample.sequencingLane
+          ? sample.run.name + "(L" + sample.sequencingLane + ")"
+          : sample.run.name,
         urls.miso.run(sample.run.name),
         urls.dimsum.run(sample.run.name)
       )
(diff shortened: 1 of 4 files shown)


ts/data/case.ts
```diff
@@ -589,7 +589,9 @@ function makeSampleTooltip(sample: Sample) {
   if (sample.run) {
     topContainer.appendChild(
       makeNameDiv(
-        sample.run.name,
+        sample.sequencingLane
+          ? sample.run.name + "(L" + sample.sequencingLane + ")"
```
mlaszloffy, line 593: This may be slightly easier to read with a space between `run name` and `(L#)`

mlaszloffy (approved):

ts/data/sample.ts
```diff
@@ -85,7 +86,13 @@ function makeNameColumn(includeRun: boolean): ColumnDefinition<Sample, void> {
       if (includeRun && sample.run) {
         const runName = sample.run.name;
         fragment.appendChild(
-          makeNameDiv(runName, urls.miso.run(runName), urls.dimsum.run(runName))
+          makeNameDiv(
+            sample.sequencingLane
+              ? runName + "(L" + sample.sequencingLane + ")"
```
djcooke, line 91: Need a space here too

djcooke (approved):

PR #74, opened 2022-09-28
[GLT-3744] Add escape option for filters

mengjenn:
As per [GLT-3744](https://jira.oicr.on.ca/browse/GLT-3744)

Because most of the dropdown menu buttons employ the same parent class, an extra param (displayTemporary) was added in order to specify whether a button should be removed if an event calls for it, or if they are "permanent" buttons. E.g. The "Items per page" would be a permanent button (should not be removed by hitting esc or clicking out), but buttons created from selecting a filter option are temporary (can be removed by hitting esc or clicking out).

### DIFF
--- a/ts/component/dropdown.ts
+++ b/ts/component/dropdown.ts
@@ -26,7 +26,6 @@ export class BasicDropdownOption implements DropdownOption {
 
 export class Dropdown {
   private dropdownContainer: HTMLElement;
-
   // displayTemporary: set to true if the created dropdown menu can be removed
   // (by user interaction or otherwise), else set to false (or undefined) if otherwise
   constructor(
@@ -54,41 +53,45 @@ export class Dropdown {
       dropdownClickout.classList.toggle("hidden");
     };
 
-    dropdownButton.onclick = toggleMenu;
+    var handleKeydown = (event: KeyboardEvent) => {
+      // only remove the button in question if it is not already hidden
+      if (event.key === "Esc" || event.key === "Escape") {
+        if (displayTemporary) {
+          this.dropdownContainer.remove();
+        } else {
+          if (
+            !dropdownMenuContainer.classList.contains("hidden") &&
+            !dropdownClickout.classList.contains("hidden")
+          ) {
+            toggleMenu();
+          }
+        }
+        document.removeEventListener("keydown", handleKeydown);
+        console.log("REMOVE ON HANDLE");
+      }
+    };
+
+    dropdownButton.onclick = () => {
+      toggleMenu();
+      if (!dropdownMenuContainer.classList.contains("hidden")) {
+        console.log("ADDED ON DROPDOWN CLICK");
+        document.addEventListener("keydown", handleKeydown);
+      }
+    };
     // close dropdown menu by clicking outside of the menu or by hitting Esc
     dropdownClickout.onclick = () => {
       if (displayTemporary && !dropdownButton.classList.contains("hidden")) {
-        dropdownButton.remove();
+        this.dropdownContainer.remove();
       }
       toggleMenu();
+      console.log("REMOVED ON CLICKOUT");
+      document.removeEventListener("keydown", handleKeydown);
     };
-    document.addEventListener(
-      "keydown",
-      (event) => {
-        // only remove the button in question if it is not already hidden
-        if (event.key === "Esc" || event.key === "Escape") {
-          const menuHidden = dropdownMenuContainer.classList.contains("hidden");
-          const clickoutHidden = dropdownClickout.classList.contains("hidden");
-          if (displayTemporary) {
-            // dropdown button is temporary, remove it and its corresponding elements
-            if (
-              !dropdownButton.classList.contains("hidden") &&
-              !menuHidden &&
-              !clickoutHidden
-            ) {
-              dropdownMenuContainer.remove();
-              dropdownButton.remove();
-            }
-          } else {
-            // button is NOT temporary, hide its corresponding elements if they are visible
-            if (!menuHidden && !clickoutHidden) {
-              toggleMenu();
-            }
-          }
-        }
-      },
-      displayTemporary ? { once: true } : false
-    );
+    // create an additional event listener for temporary buttons
+    if (displayTemporary) {
+      document.addEventListener("keydown", handleKeydown);
+      console.log("ADD ON TEMP BUTTON");
+    }
 
     DropdownOptions.forEach((option) => {
       const li = document.createElement("li");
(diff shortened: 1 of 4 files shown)


ts/component/dropdown.ts
```diff
@@ -44,16 +47,39 @@ export class Dropdown {
...
+        const clickoutHidden = dropdownClickout.classList.contains("hidden");
+        if (displayTemporary) {
+          // dropdown button is temporary, hide its corresponding elements if they are visible
+          if (
+            !dropdownButton.classList.contains("hidden") &&
+            !menuHidden &&
+            !clickoutHidden
+          ) {
+            toggleDropdownButton();
+          }
```
djcooke, line 74: Elements should be removed from DOM entirely instead of hiding them. Can just remove `dropdownContainer` to get rid of the entire thing

ts/component/text-input.ts
```diff
@@ -42,8 +42,19 @@ export class TextInput {
...
+      this.container.classList.remove("ring-2");
+      textInputClickout.classList.toggle("hidden");
+    };
     submitIcon.onclick = submitTextInput;
-    textInputClickout.onclick = submitTextInput;
+    // Close text input field by clicking outside or hitting esc
+    textInputClickout.onclick = toggleTextInput;
+    textInputClickout.addEventListener("keypress", (event) => {
+      if (event.key === "Escape" || event.key === "Esc") {
+        toggleTextInput();
```
djcooke, line 55: Should be removed instead of hiding

ts/component/dropdown.ts
```diff
@@ -44,16 +47,39 @@ export class Dropdown {
...
+        dropdownButton.classList.toggle("hidden");
+      }
+      toggleMenu();
     };
     dropdownButton.onclick = toggleMenu;
     dropdownButton.innerHTML = makeDisplayText(displayLabel, defaultOption);
-    dropdownClickout.onclick = defaultOption ? toggleMenu : invalidInput;
+    // close dropdown menu by clicking outside of the menu or by hitting Esc
+    dropdownClickout.onclick = toggleDropdownButton;
+    document.addEventListener("keydown", (event) => {
```
djcooke, line 61: These listeners will outlive their usefulness and continue to accumulate. Make sure to remove the listener when removing the dropdown

ts/component/text-input.ts
```diff
@@ -59,6 +58,26 @@ export class TextInput {
...
+
+    document.addEventListener(
+      "keydown",
+      (event) => {
+        // only remove the text input box in question if it is not already hidden
+        if (event.key === "Esc" || event.key === "Escape") {
+          const containerHidden = this.container.classList.contains("hidden");
+          const textFieldHidden = this.textField.classList.contains("hidden");
+          // remove text input field corresponding elements if they are visible
+          if (!containerHidden && !textFieldHidden) {
```
djcooke, line 74: No need to check if it's hidden anymore

djcooke (commented):
Need to remove the listener in `text-input.ts` `submitTextInput` as well

djcooke:
Were you going to add the fix for the z-index issue to this PR?

ts/util/html-utils.ts
```diff
@@ -98,7 +98,7 @@ export function addIconButton(container: HTMLElement, iconName: string) {
 export function makeClickout() {
   const clickout = document.createElement("button");
   clickout.className =
-    "bg-transparent fixed inset-0 w-full h-full cursor-default hidden z-10";
+    "bg-transparent fixed inset-0 w-full h-full cursor-default hidden";
```
djcooke, line 101: `z-10` is removed from here, but then re-added later in each place that the function is called?

ts/component/text-input.ts
```diff
@@ -20,14 +20,22 @@ export class TextInput {
...
+      "ring-0 border-0 outline-0 rounded-sm relative px-1 min-w-[150px] z-40";
     this.textField.setAttribute("size", "10");
-    submitIcon.classList.add("text-black", "hover:text-green", "relative");
+    submitIcon.classList.add(
+      "text-black",
+      "hover:text-green",
+      "relative",
+      "z-40"
+    );
+    label.classList.add("relative", "z-40");
```
djcooke, line 37: This is much better, but there is still an issue if you click part of the component that is not covered by these individual elements. This could easily happen by accident, like if you click just above or below the submit button, but still within the component. Nothing should happen if you click there, but currently the clickout is over-top of it, so it cancels the input. The solution we discussed previously of adding an additional container around the component and clickout would fix this and also reduce the number of places that z-index needs to be set:

```
<div> <!-- new outer container -->
  <div class="... z-40> <!-- existing container -->
    <span .../>
    <input .../>
    <i .../>
  </div>
  <button ... class="... z-10" .../> <!-- clickout -->
</div>
```

ts/component/text-input.ts
```diff
@@ -84,10 +70,20 @@ export class TextInput {
 
     document.addEventListener("keydown", handleEsc);
 
-    this.container.append(label);
-    this.container.appendChild(this.textField);
+    this.container.className = "inline-block flex-auto items-center";
```
djcooke, line 73: `items-center` is for flex containers, so has no effect here

ts/component/text-input.ts
```diff
@@ -84,10 +70,20 @@ export class TextInput {
...
+    var textInputContainer = document.createElement("div");
+    textInputContainer.className =
+      "font-inter font-medium text-12 text-black px-2 py-1 rounded-md ring-2 ring-offset-1 ring-red inline-block flex-auto items-center space-x-2 relative z-40";
```
djcooke, line 76: I realize you probably didn't add these, but
* `items-center` is for flex containers, so has no effect here
* `flex-auto` is for flex contents, so has no effect here

ts/component/text-input.ts
```diff
@@ -84,10 +70,20 @@ export class TextInput {
...
+    this.textField.className =
+      "ring-0 border-0 outline-0 rounded-sm px-1 min-w-[150px]";
+    this.textField.setAttribute("size", "10");
+    submitIcon.classList.add("text-black", "hover:text-green");
+
+    textInputContainer.append(label);
+    textInputContainer.appendChild(this.textField);
+    textInputContainer.appendChild(submitIcon);
+    this.container.append(textInputContainer);
     this.container.append(this.datalist);
```
djcooke, line 86: `datalist` has to go inside the `textInputContainer` too - auto-complete is not working currently

djcooke (commented):
`styleValidity` isn't working currently either - the outline stays red even when there is valid text entered

djcooke (approved):

PR #76, opened 2022-09-30
[GLT-3779] create notifications only when metrics are available

djcooke:


mengjenn (approved):

ts/data/notification.ts
```diff
@@ -13,27 +23,85 @@ export const notificationDefinition: TableDefinition<Run, void> = {
...
+        addParentContents(notification, fragment) {
+          switch (notification.metricCategory) {
+            case "LIBRARY_QUALIFICATION":
+              addText(fragment, "Library Qualification");
+              break;
+            case "FULL_DEPTH_SEQUENCING":
+              addText(fragment, "Full-Depth Sequencing");
+              break;
+            default:
+              addText(fragment, "unknwon/error");
```
mlaszloffy, line 60: unknwon -> unknown

mlaszloffy (approved):

ts/data/notification.ts
djcooke: fixed, thanks

PR #77, opened 2022-10-05
GLT-3769 Specify filters in url

mengjenn:
As per [GLT-3769](https://jira.oicr.on.ca/browse/GLT-3769)

Currently, sorting and items per page options will not appear in the url, but this functionality can be added if desired.

### DIFF
--- a/ts/component/table-builder.ts
+++ b/ts/component/table-builder.ts
@@ -102,6 +102,7 @@ export class TableBuilder<ParentType, ChildType> {
   definition: TableDefinition<ParentType, ChildType>;
   container: HTMLElement;
   table?: HTMLTableElement;
+  thead?: HTMLTableSectionElement;
   pageDescription?: HTMLSpanElement;
   pageLeftButton?: HTMLButtonElement;
   pageRightButton?: HTMLButtonElement;
@@ -175,6 +176,7 @@ export class TableBuilder<ParentType, ChildType> {
     });
 
     this.load();
+    this.setupScrollListener();
     this.reload();
   }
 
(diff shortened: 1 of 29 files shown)


djcooke (commented):
Works great, but needs some refactoring to sort out responsibilities.

Another responsibility issue is the setting of URL parameters. Right now, it's happening in `dropdown.ts` and `text-input.ts`. I suggested moving most of that code into `table-builder.ts`, but the table-builder shouldn't actually be manipulating the URL parameters for the same reason that it shouldn't be reading them. I think when constructing the table-builder, you should pass it an optional "onFilterChange" function. That way, the table-builder can call that whenever the filters change, and the page script can define that function to update the URL parameters

ts/component/table-builder.ts
```diff
@@ -135,28 +138,28 @@ export class TableBuilder<ParentType, ChildType> {
...
-            new AcceptedFilter(key, key.toUpperCase(), value, () =>
-              this.reload()
-            )
-          );
-        }
+
+    if (filterParams) {
+      // create accepted filter if there is a matching key in table definition
+      filterParams.forEach((p) => {
+        this.definition.filters?.find((f) => {
```
djcooke, line 145: Could use `forEach` instead of `find` since you're not doing anything with the return value

ts/util/urls.ts
```diff
@@ -1,9 +1,14 @@
 import { siteConfig } from "./site-config";
 import { toTitleCase } from "./html-utils";
+import { Pair } from "./pair";
 
 const restBaseUrl = "/rest";
 const misoBaseUrl = `${siteConfig.misoUrl}/miso`;
 
+export interface Param extends Pair<string, string> {
```
djcooke, line 8: Is this used for something?

djcooke (approved):

mlaszloffy (approved):

PR #78, opened 2022-10-05
[GLT-3778] Show Run-Library Lane Number on Run Details Page

mengjenn:
As per [GLT-3778](https://jira.oicr.on.ca/browse/GLT-3778)

### DIFF
--- a/ts/data/sample.ts
+++ b/ts/data/sample.ts
@@ -88,12 +88,21 @@ function makeQcStatusColumn(
   };
 }
 
-function makeNameColumn(includeRun: boolean): ColumnDefinition<Sample, void> {
+function makeNameColumn(
+  includeRun: boolean,
+  includeLane?: boolean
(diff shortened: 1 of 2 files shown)


ts/data/sample.ts
```diff
@@ -88,12 +88,21 @@ function makeQcStatusColumn(
   };
 }
 
-function makeNameColumn(includeRun: boolean): ColumnDefinition<Sample, void> {
+function makeNameColumn(
+  includeRun: boolean,
+  includeLane?: boolean
```
djcooke, line 93: This extra parameter is unnecessary since it should always be the opposite of `includeRun`. If both parameters were true, the lane would be shown twice. There won't be any case where we don't want to show the lane if there is one, so we can just show the lane with the sample name `if (!includeRun && sample.sequencingLane)`

djcooke (approved):

mlaszloffy (approved):

PR #88, opened 2022-11-07
[GLT-3740] Tab interface for details page

mengjenn:
Create tab bar for detail pages to navigate between different tables. Selected filters will be applied to every table. Upon selecting a tab, the old table will be destroyed and replaced with the new table. To facilitate this, the details template page was cleaned up and replaced with a single table container. Additionally, filtering options were added to sample and requisition.

As per [GLT-3740](https://jira.oicr.on.ca/browse/GLT-3740)


ts/component/tab-bar-builder.ts
```diff
@@ -0,0 +1,129 @@
+import { Pair } from "../util/pair";
+import { getSearchParams, updateUrlParams } from "../util/urls";
+import { TableBuilder } from "./table-builder";
+
+class Tab {
+  tabContainer: HTMLButtonElement;
```
djcooke, line 6: Rename `tabButton`

ts/component/tab-bar-builder.ts
```diff
@@ -0,0 +1,129 @@
...
+  table: Pair<any, string>;
+  selected: boolean = false;
+  onSelect: (table: Pair<any, string>) => void;
```
djcooke, line 9: Remove `table` field and argument from `onSelect` function. The only time `table` is used is when calling `onSelect`, but these are both provided in the constructor, so the caller already has access to the table to use within `onSelect`. The tab bar could also be reused for other purposes not involving tables

ts/component/tab-bar-builder.ts
```diff
@@ -0,0 +1,129 @@
...
+        "ring-2",
+        "ring-green-200",
+        "ring-offset-1"
+      );
+    } else {
+      this.tabContainer.classList.replace("text-green-200", "text-black");
+      this.tabContainer.classList.remove(
+        "ring-2",
+        "ring-green-200",
+        "ring-offset-1"
```
djcooke, line 48: `ring-green-200` and `ring-offset-1` are re-added whether selected or not. Can add them when creating the button and then never remove them

ts/component/tab-bar-builder.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.tabContainer = button;
+    this.styleButton();
+  }
+
+  // modifies the tab styling according to whether or not it has been selected
+  public styleButton() {
+    if (this.selected) {
+      this.tabContainer.classList.replace("text-black", "text-green-200");
+      this.tabContainer.classList.add(
+        "ring-2",
```
djcooke, line 39: Can use `replace` to switch between `ring-2` and `hover:ring-2`

ts/component/tab-bar-builder.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (tab.table.value === table.value) {
+        tab.selected = true;
+        // destroy current table and construct new table
+        this.tableContainer.innerHTML = "";
+        new TableBuilder(
+          table.key,
+          this.tableContainer.id,
+          getSearchParams(),
+          updateUrlParams
+        ).build();
```
djcooke, line 122: Move this logic to `detail.ts`. Instead of `tables`, the `TabBar` constructor can take a `Pair<string, () => void>[]`. This way `TabBar` is only responsible for its own behaviour and calling the function defined by its owner. In our current use case, that is rebuilding a table, but it could be used for other things in the future

mengjenn: I've modified the tables array to be Pair<Pair<any, string>, () => void>, as when rebuilding the table we'll need the table definition as well. So for example, for the "Cases" table, its pair would look like Pair(Pair(caseDefinition, "Cases"), reload).

ts/data/case.ts
```diff
@@ -98,51 +103,49 @@ export interface Case {
   latestActivityDate: string;
 }
 
+export var caseFilters: FilterDefinition[] = [
```
djcooke, line 106: This is now ending up `undefined` when `sample.ts` attempts to access it. This is because of a circular dependency between `case.ts` and `sample.ts`. I think the best way to sort it out would be to create a new `table-components.ts` module that holds `caseFilters`. Other similar definitions that are shared between multiple table definitions can be moved here as well

ts/details.ts
```diff
@@ -11,40 +11,68 @@ import {
...
+const tableContainerId = "tableContainer";
 
-new TableBuilder(
-  getFullDepthSequencingsDefinition(urls.rest.fullDepthSequencings, true),
-  "fullDepthSequencingsTableContainer"
-).build();
+// each pair consists of (1) the information required to build the table (definition
+// and title), and (2) its reload function upon selection
+const tables = [
+  new Pair(new Pair(caseDefinition, "Cases"), reload),
```
djcooke, line 24: This still has the tab-bar dealing with tables, which it shouldn't. Instead, you can add the table definition as a parameter to the reload function, and these tab definitions could be `Pair<string, () => void`. Here, this would look something like: `new Pair("Cases", () => reload(caseDefinition))`

ts/details.ts
```diff
@@ -76,3 +69,22 @@ function reload() {
...
+//         tableContainerId,
+//         getSearchParams(),
+//         updateUrlParams
+//       ).build();
+//     } else {
+//       tab.selected = false;
+//     }
+//     tab.styleButton();
+//   });
+// }
```
djcooke, line 90: Remove commented code

ts/details.ts
```diff
@@ -57,15 +50,15 @@ const tabBar = new TabBar(
 
 tabBar.build();
 
-// reload
-function reload() {
-  tabBar.tables.forEach((tbl, idx) => {
+// reload: destroy current table and build new table
+function reload(definition: TableDefinition<any, any>) {
```
djcooke, line 54: This function should only destroy and re-create the table. Tab-switching logic should be in `tab-bar-builder.ts`

ts/component/tab-bar-builder.ts
```diff
@@ -40,14 +40,14 @@ class Tab {
...
   tabBarContainer: HTMLElement;
   tableContainer: HTMLElement;
   tabs: Tab[] = [];
 
   constructor(
-    tables: Pair<Pair<any, string>, () => void>[],
+    elements: Pair<string, () => void>[],
     defaultTable: string,
     tabBarContainerId: string,
     tableContainerId: string
```
djcooke, line 53: This is not used for anything (and shouldn't be known to `TabBar`)

djcooke (commented):
Make sure all logic is all where it belongs:
* `tab-bar-builder.ts`: rendering the tab bar, switching tabs, and calling (but not defining) the action handler. Should not have any knowledge of tables
* `detail.ts`: destroying and rebuilding the tables. Should not require any knowledge of how the `TabBar` works or access to its internals

djcooke (approved):

PR #99, opened 2022-12-13
[GLT-3820] added Omissions list page

djcooke:


mlaszloffy (approved):

src/main/resources/templates/omissions.html
```diff
@@ -0,0 +1,24 @@
...
+<head>
+  <meta charset="UTF-8" />
+  <title>Omissions</title>
+</head>
+
+<body>
+  <div layout:fragment="content">
+
+    <h1 class="text-green-200 font-sarabun font-light font text-32">
+      Omissions <i id="omissionsInfo" class="fa-solid fa-circle-question text-black"></i>
```
avarsava, line 14: empty italics tag?

avarsava (approved):

src/main/resources/templates/omissions.html
djcooke: The `i` tag originally meant "italics," but has been expanded. It's the tag Font Awesome recommends using for icons, and the classes cause Font Awesome icons to be added (in this case, a solid circle with question mark)

PR #100, opened 2022-12-14
[GLT-3824] fix determining JIRA issue state from comments

djcooke:


mlaszloffy (approved):

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
+  // Use IntelliSense to learn about possible attributes.
```
avarsava, line 2: comments in json aren't part of the standard so they're generally not supported and cause I/O failures, does vscode accept them?

djcooke: VSCode generated the comment, and accepts them, as I was able to run the command that this creates in VSCode

avarsava (approved):

PR #107, opened 2023-03-06
[GLT-3867] indicate items with different assays

djcooke:


avarsava (approved):

ts/data/case.ts
```diff
@@ -649,7 +660,7 @@ function addTooltipRow(
   container: HTMLElement, // must be a two column grid
   label: string,
   name: string,
-  misoUrl: string,
+  misoUrl?: string,
```
hyeonjeongkwon, line 663: Why would miso url be optional?

djcooke: This function is for adding a row (basically a key and value pair) to the tooltip. Previously included rows all had MISO links, but this PR adds the assay without a MISO link

hyeonjeongkwon (approved):

PR #108, opened 2023-03-07
[GLT-3854] Generates Project Summary

hyeonjeongkwon:


src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/PendingState.java
```diff
@@ -304,23 +304,23 @@ public Predicate<Sample> samplePredicate(MetricCategory requestCategory) {
...
     return true;
   }
 
-  protected boolean qualifyRequisition(Requisition requisition) {
+  public boolean qualifyRequisition(Requisition requisition) {
     return true;
   }
 
-  protected Function<Case, Stream<Sample>> getFilteredSamples(MetricCategory requestCategory) {
+  public Function<Case, Stream<Sample>> getFilteredSamples(MetricCategory requestCategory) {
```
djcooke, line 323: Does this need to be `public`?

src/main/java/ca/on/oicr/gsi/dimsum/data/CaseData.java
```diff
@@ -23,10 +23,12 @@ public class CaseData {
   private final Set<String> requisitionNames;
   private final Set<String> projectsNames;
   private final Set<String> donorNames;
+  private final Map<String, ProjectSummary> projectsBySummary;
```
djcooke, line 26: Rename `projectSummariesByName` (values are `ProjectSummary`, key is the project name)

src/main/java/ca/on/oicr/gsi/dimsum/data/ProjectSummary.java
```diff
@@ -0,0 +1,53 @@
...
+
+import static java.util.Objects.requireNonNull;
+import java.util.Collections;
+import java.util.Map;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class ProjectSummary {
+  private final String name;
+  private final Map<String, Integer> counts;
```
djcooke, line 11: We'll want a stricter data model to enforce a specific set of keys. This also will not convert to a nice JSON representation. Ideal implementation will use specific `int` fields such as `receiptPendingQcCount`, `receiptCompletedCount`, `extractionPendingCount`, etc.

You could add a `addCounts(ProjectSummary.Builder)` method to `ProjectSummary.Builder` to make it easy to combine counts. Then in `CaseLoader`, you'd use a `ProjectSummary.Builder caseSummary` to add up the case counts, and `projectSummary.addCounts(caseSummary)` to push those counts into the project

src/test/java/ca/on/oicr/gsi/dimsum/CaseLoaderTest.java
```diff
@@ -242,6 +250,491 @@ public void testLoad() throws Exception {
...
+    assertEquals(1, projectsBySummary.get("PRO21").getCounts().get("LIBRARY_PREP_COMPLETE"));
+
+    // PRO22
+    assertEquals(1, projectsBySummary.get("PRO22").getCounts().get("TOTAL_TESTS"));
+    assertEquals(1, projectsBySummary.get("PRO22").getCounts().get("RECEIPT_PEND_QC"));
+
+  }
+
+
+  private static Case makeCase0() {
```
djcooke, line 354: Rather than duplicating these from `CaseFilterTest`, move them to a separate `MockCases` class that you can call from both of these test classes

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/PendingState.java
hyeonjeongkwon: i think some of the preliminary changes that I was trying out before rebase also got pushed. I will make it `protected`. Sorry for the confusion

src/main/java/ca/on/oicr/gsi/dimsum/data/ProjectSummary.java
hyeonjeongkwon: Would adding the counts within the builder be feasible? From my understanding of your suggestion I was thinking of something like `this.totalTestCount` += `caseSummary.totalTestCount`, but this would modify the field of existing `ProjectSummary` object.

djcooke: Yes, counts would all have to be added within the builder. Sorry I wasn't clear, but in my example of `projectSummary.addCounts(caseSummary)`, `projectSummary` and `caseSummary` were both meant to be builders

hyeonjeongkwon: Ah I think I see what you mean. So just to clarify, we would keep track of the builder objects (add the counts if necessary) and push those to the `ProjectSummariesbyName` when we iterate through all the cases?

djcooke: `CaseData.projectSummariesByName` would still be a `Map<String, ProjectSummary>`.

In `CaseLoader#calculateProjectSummaries`, you would have a `Map<String, ProjectSummary.Builder>` to track the project summaries temporarily. Each case would be converted to a `ProjectSummary.Builder` which I'm called `caseSummary`, then for each project in the case, you'd get the project summary from the map (or add new if necessary), and add the caseCounts to it. Once all cases are handled, you'd build all of the `ProjectSummary`s

hyeonjeongkwon: Thanks for clarification

src/main/java/ca/on/oicr/gsi/dimsum/data/CaseData.java
```diff
@@ -91,4 +95,8 @@ public Set<String> getRunNames() {
   public Collection<RunAndLibraries> getRunsAndLibraries() {
     return runsByName.values();
   }
+
+  public Map<String, ProjectSummary> getprojectSummariesByName() {
```
djcooke, line 99: Missing capital "P"

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -697,4 +699,112 @@ private <T> List<T> loadFromJsonArrayFile(FileReader fileReader, ParseFunction<J
     return loaded;
   }
 
+  protected static Map<String, ProjectSummary> calculateProjectSummaries(List<Case> cases) {
+    Map<String, ProjectSummary.Builder> tempprojectSummariesByName = new HashMap<>();
+    Map<String, ProjectSummary> projectSummariesByName = new HashMap<>();
+
+    for (Case kase : cases) {
+      ProjectSummary.Builder builder =
```
djcooke, line 707: rename `builder` -> `caseSummary` for clarity

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -697,4 +699,112 @@ private <T> List<T> loadFromJsonArrayFile(FileReader fileReader, ParseFunction<J
...
+
+      // add the counts to each project in the case if the project exists in the
+      // projectSummariesByName
+      for (Project project : kase.getProjects()) {
+        builder.name(project.getName());
+        if (tempprojectSummariesByName.containsKey(project.getName())
+            && !tempprojectSummariesByName.isEmpty()) {
+          tempprojectSummariesByName.get(project.getName()).addCounts(builder);
+        } else {
+          tempprojectSummariesByName.put(project.getName(), builder);
```
djcooke, line 798: You'll need to create a new `ProjectSummary.Builder` and use `addCounts` to copy in the counts here. Otherwise, this will happen:

1. Case 1 involves PROJ1 and PROJ2, so in your map, "PROJ1" and "PROJ2" will both be set to the same object
2. Case 2 involves only PROJ1, so you use `addCounts` (above) to copy the case summary into "PROJ1"
3. Since "PROJ2" is the same object as "PROJ1", "PROJ2" now includes the counts for Case 2

changes/add_project_summary.md
```diff
@@ -0,0 +1 @@
+Project Summary generates counts of items at each QC gate that are pending, pending QC, completed
```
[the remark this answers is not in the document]
djcooke: Good point. It's not useful to have a changelog message for this

src/main/java/ca/on/oicr/gsi/dimsum/data/ProjectSummary.java
```diff
@@ -0,0 +1,339 @@
...
+      // draft report
+      this.draftReportPendingCount += builder.draftReportPendingCount;
+      this.draftReportCompletedCount += builder.draftReportCompletedCount;
+      // final report
+      this.finalReportPendingCount += builder.finalReportPendingCount;
+      this.finalReportCompletedCount += builder.finalReportCompletedCount;
+    }
+
+    public ProjectSummary build() {
+      return new ProjectSummary(this);
```
avarsava, line 336: `Builder.build()` and `new ProjectSummary(builder)` are both public, though neither seems to ever be called? Which is preferred?

avarsava:
I know these will all be squashed at merge, but writing descriptive commit messages is still a good habit to be in :)

avarsava (approved):

changes/add_project_summary.md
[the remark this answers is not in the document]
hyeonjeongkwon: I'll delete the changelog message

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -700,107 +700,109 @@ private <T> List<T> loadFromJsonArrayFile(FileReader fileReader, ParseFunction<J
...
       }
 
       // add the counts to each project in the case if the project exists in the
       // projectSummariesByName
       for (Project project : kase.getProjects()) {
-        builder.name(project.getName());
-        if (tempprojectSummariesByName.containsKey(project.getName())
-            && !tempprojectSummariesByName.isEmpty()) {
-          tempprojectSummariesByName.get(project.getName()).addCounts(builder);
+        caseSummary.name(project.getName());
```
djcooke, line 793: Don't need to set a name on the case summary

djcooke (approved):

PR #109, opened 2023-03-08
[GLT-3871] Fix Dimsum Naming Consistency

hyeonjeongkwon:


djcooke (approved):

changes/fix_nameConsistency.md
```diff
@@ -0,0 +1 @@
+Project name should all be "Dimsum" instead of "Dim Sum"
```
avarsava, line 1: This reads more like an outstanding issue than a description of a change, and it's ambiguous. Are they all 'Dimsum' or aren't they? I would recommend something like 'Changed references to previous name, 'Dim Sum', to 'Dimsum''.

avarsava (commented):
You may want to rename `DimSumApplication`'s file and class name too

avarsava (approved):

PR #111, opened 2023-03-08
[GLT-3865] added requisition stop reason

djcooke:
`stopped` was previously on `Case` because all involved requisitions needed to be stopped for the case to be stopped. Now that a case is based on a single requisition, this was redundant, so I removed it


avarsava:
Tooltip is the only way to see the stop reason? Will new users know to hover over 'CASE STOPPED'?

djcooke:
> Tooltip is the only way to see the stop reason? Will new users know to (/ be able to discover) hover over 'CASE STOPPED'?

It's not important enough to take up any additional space in the table, and the worst case if users don't find it is that they go to MISO or ask someone why the case was stopped

avarsava (approved):

hyeonjeongkwon (approved):

PR #112, opened 2023-03-09
[GLT-3873] Remove turn-around time for stopped cases

hyeonjeongkwon:


avarsava (approved):

ts/data/case.ts
```diff
@@ -704,5 +704,5 @@ function getElapsedMessage(kase: Case) {
   const startDate = new Date(kase.startDate);
   const milliDiff = endDate.getTime() - startDate.getTime();
   const dayDiff = Math.ceil(milliDiff / dayMillis);
-  return `(${message} ${dayDiff} days)`;
+  return !kase.requisition.stopped ? `(${message} ${dayDiff} days)` : "";
```
djcooke, line 707: This is still constructing the message that's not being used, and adding unnecessary empty DOM elements. Please change to not create `elapsedDiv` instead

hyeonjeongkwon: Will do. I just have a quick follow up question. Will the case be stopped (`kase.requisition.stopped == true`) when the case is completed?

djcooke: No, "stopped" is like cancelling the case

djcooke (approved):

PR #115, opened 2023-03-14
[GLT-3859] added test table

hyeonjeongkwon:


src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -307,6 +309,21 @@ public TableData<ProjectSummary> getProjects(int pageSize, int pageNumber,
...
   }
 
+  public TableData<TestTableView> getTestTableViews(int pageSize, int pageNumber,
+      TestTableViewSort sort, boolean descending,
+      CaseFilter baseFilter, Collection<CaseFilter> filters) {
+    List<Case> cases = getCases(baseFilter);
+    TableData<TestTableView> data = new TableData<>();
+    data.setTotalCount(
+        cases.stream().flatMap(kase -> kase.getTests().stream()).distinct().count());
+    List<TestTableView> testTableViews = filterTestTableViews(cases, filters).distinct().toList();
```
djcooke, line 319: Is there a reason for distinct? I don't think there should ever be duplicates

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CaseFilterKey.java
```diff
@@ -66,6 +67,10 @@ public Function<String, Predicate<Requisition>> requisitionPredicate() {
     return string -> requisition -> true;
   }
 
+  public Function<String, Predicate<TestTableView>> testTableViewPredicate() {
+    return string -> testTableViewPredicate -> true;
```
djcooke, line 71: You'll have to override this for `PENDING`, similar to `testPredicate`. You can add methods to `PendingState` to reuse the existing code for handling `Test`s. e.g.

```
public Predicate<TestTableView> testTableViewPredicate() {
  return view -> testPredicate.test(view.getTest());
}
```

ts/data/test.ts
```diff
@@ -0,0 +1,371 @@
...
+  receipts: Sample[];
+}
+
+export const testDefinition: TableDefinition<TestTableView, void> = {
+  queryUrl: urls.rest.tests,
+  defaultSort: {
+    columnTitle: "Test",
+    descending: true,
+    type: "text",
+  },
```
djcooke, line 51: I think we need to add a "Latest Activity" column to use for sorting because that will be much more useful, and should be the default sort (descending)

ts/data/test.ts
```diff
@@ -0,0 +1,371 @@
...
+      },
+    },
+  ],
+};
+
+function handleNaSamplePhase(
+  testTableView: TestTableView,
+  samples: Sample[],
+  fragment: DocumentFragment
+) {
```
djcooke, line 328: Only difference from the function in `case.ts` is it takes a `Case` instead of `TestTableView`, but both only really need the `Requisition`. Change the original function to take a requisition instead and use that in both places.

Similar situation for `addSampleIcons` and `getSamplePhaseHighlight` below

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
hyeonjeongkwon: I'll remove them!

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CaseFilterKey.java
hyeonjeongkwon: What is the difference between the code above and using `return view -> qualifytest(view.getTest())`?

djcooke: They'd have the same effect, though your suggestion is slightly better since it's more direct

hyeonjeongkwon: @djcooke Also, wouldn't this cause an error when trying to filter by `PendingState` that uses requisition (ex. final report)? Should I add `qualifytest` for those states and return false?

src/main/java/ca/on/oicr/gsi/dimsum/data/TestTableView.java
```diff
@@ -21,18 +20,20 @@ public class TestTableView {
...
     this.donor = requireNonNull(kase.getDonor());
-    this.projects = unmodifiableSet(kase.getProjects());
+    this.projects = kase.getProjects();
     this.assay = requireNonNull(kase.getAssay());
     this.tissueOrigin = requireNonNull(kase.getTissueOrigin());
     this.tissueType = requireNonNull(kase.getTissueType());
     this.timepoint = kase.getTimepoint();
-    this.receipts = unmodifiableList(kase.getReceipts());
+    this.receipts = kase.getReceipts();
+    this.latestActivityDate = kase.getLatestActivityDate();
```
djcooke, line 36: `Test` has its own `latestActivityDate` already

djcooke (approved):

avarsava (approved):

PR #118, opened 2023-03-16
[GLT-3860] New case filters

hyeonjeongkwon:


src/main/java/ca/on/oicr/gsi/dimsum/FrontEndConfig.java
```diff
@@ -28,6 +30,11 @@ public class FrontEndConfig {
   private final List<String> pendingStates =
       Stream.of(PendingState.values()).map(PendingState::getLabel).toList();
 
+  private final List<String> completedGates =
+      Stream.of(CompletedGate.values()).map(CompletedGate::getLabel).toList();
+
+  private final List<String> stopStatus = Arrays.asList("Yes", "No");
```
djcooke, line 36: These don't need to be in the config - the values can be hard-coded in the typescript instead of hard-coding them here

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -158,6 +159,13 @@ public Set<String> getMatchingRunNames(String prefix) {
         .collect(Collectors.toSet());
   }
 
+  public Set<String> getMatchingTestNames(String prefix) {
+    Set<Test> tests = caseData.getCases().stream().flatMap(kase -> kase.getTests().stream())
+        .filter(s -> s.getName().toLowerCase().startsWith(prefix.toLowerCase()))
+        .collect(Collectors.toSet());
+    return tests.stream().map(test -> test.getName()).collect(Collectors.toSet());
+  }
```
djcooke, line 167: This could potentially get slow as more and more cases are added to Dimsum, and it gets called rapidly as someone is typing. Collect the test names once in `CaseLoader` and store in `CaseData` instead

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CompletedGate.java
```diff
@@ -0,0 +1,239 @@
...
+  public Predicate<TestTableView> testTableViewPredicate() {
+    return view -> qualifyTest(view.getTest());
+  }
+
+  public Predicate<Sample> samplePredicate(MetricCategory requestCategory) {
+    return sample -> qualifySample(sample, requestCategory);
+  }
+
+  public boolean qualifyCase(Case kase) {
+    return kase.getTests().stream().anyMatch(test -> qualifyTest(test));
```
djcooke, line 179: All test must be completed for the case to be completed

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CaseFilterKey.java
```diff
@@ -47,7 +47,34 @@ public Function<String, Predicate<TestTableView>> testTableViewPredicate() {
   PROJECT(string -> kase -> kase.getProjects().stream()
       .anyMatch(project -> project.getName().equalsIgnoreCase(string))),
   REQUISITION(string -> kase -> kase.getRequisition().getName().toLowerCase().startsWith(string.toLowerCase())),
-  REQUISITION_ID(string -> kase -> kase.getRequisition().getId() == Long.parseLong(string));
+  REQUISITION_ID(string -> kase -> kase.getRequisition().getId() == Long.parseLong(string)),
+  TEST(string -> kase -> kase.getTests().stream().anyMatch(test -> test.getName().equalsIgnoreCase(string))),
```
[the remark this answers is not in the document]
hyeonjeongkwon: Just to confirm, is this to filter the tests with specific name in `Test` page?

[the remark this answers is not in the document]
djcooke: Yeah. For example, say we have a case with 2 tests - "Tumour WG" and "Normal WG" and filter by "Test: Tumour WG"
* If we are looking at the Cases table, we should see this case because one of its tests matches
* If we are looking at the Tests table, we should only see the "Tumour WG" test
  * currently, the "Normal WG" test would also be included 
* If we are looking at the Extractions table, we should only see the extractions for the "Tumour WG" test
  * currently, the extractions for the "Normal WG" test would also be included

avarsava (approved):

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CompletedGate.java
```diff
@@ -107,8 +108,13 @@ public boolean qualifyCase(Case kase) {
 
     @Override
     public boolean qualifyRequisition(Requisition requisition) {
-      return requisition.getInformaticsReviews().stream().allMatch(reqQc -> reqQc.isQcPassed());
-    }    
+      RequisitionQc reqQc = requisition.getInformaticsReviews().stream().max(Comparator.comparing(RequisitionQc::getQcDate)).orElse(null);
+      if (reqQc != null && reqQc.isQcPassed()) {
+        return true;
+      }
+      return false;
```
djcooke, line 115: This looks good now, but qualifyCase needs to do the same. Ideally, just modify `Helpers#isCompletedRequisitionQc` to do this and take a `Requisition` instead of a `Case`, then you can use that in all 6 places

PR #120, opened 2023-03-22
[GLT-3856] Project Summary Table

hyeonjeongkwon:


src/main/java/ca/on/oicr/gsi/dimsum/data/ProjectSummaryField.java
```diff
@@ -0,0 +1,58 @@
...
+public class ProjectSummaryField {
+  private int count;
+  private String filterKey;
+  private String filterValue;
+
+  private ProjectSummaryField(Builder builder) {
+    this.count = requireNonNull(builder.count);
+    this.filterKey = requireNonNull(builder.filterKey);
+    this.filterValue = requireNonNull(builder.filterValue);
+  }
```
djcooke, line 17: Since there are only 3 fields, it would be simpler to get rid of the builder and instead have a public constructor that takes all 3 values

ts/component/table-builder.ts
djcooke, line 1: `showLoaded` is currently throwing an error. Everything in there currently probably needs skipped if the controls are disabled, similar to `showLoading`

src/main/resources/templates/project-detail.html
```diff
@@ -0,0 +1,28 @@
...
+  <div layout:fragment="content">
+
+    <div class="flex space-x-2 items-center">
+      <h1 class="text-green-200 font-sarabun font-light font text-32" th:text="${title}">Details</h1>
+      <a th:if="${titleLink != null}" th:href="${titleLink}" target="_blank" rel="noopener noreferrer">
+        <img src="/img/miso_logo.svg" class="w-6" alt="View in MISO" title="View in MISO">
+      </a>
+    </div>
+    <div id="projectSummaryTableContainer" class="px-2"></div>
+    <div id="tabBarContainer" class="px-2"></div>
```
djcooke, line 20: Remove `px-2` from both of these so that they line up with the other elements on the page

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -325,73 +324,73 @@ public TableData<ProjectSummaryRow> getProjectSummaryRows(String name) {
     ProjectSummaryRow pendingWork = new ProjectSummaryRow.Builder()
         .title("Pending Work")
         .extraction(
-            new ProjectSummaryField.Builder().count(projectSummary.getExtractionPendingCount())
-                .filterKey("PENDING").filterValue("Extraction").build())
+            new ProjectSummaryField(projectSummary.getExtractionPendingCount(),
+                "PENDING", "Extraction"))
```
djcooke, line 328: Use the enums instead of duplicating the Strings. This way, it's easy to find usage, rename things in the future, etc.

"PENDING" -> `CaseFilterKey.PENDING.name()`
"Extraction" -> `PendingState.EXTRACTION.getLabel()`

ts/component/table-builder.ts
```diff
@@ -605,13 +603,17 @@ export class TableBuilder<ParentType, ChildType> {
     }
     const data = await response.json();
     this.load(data.items);
-    this.showLoaded(data);
+    if (!this.definition.disablePageControls) {
+      this.showLoaded(data);
+    }
```
djcooke, line 608: Move this check to inside `showLoaded`. Similar to `showLoading`, there will likely be parts of this that should be called in the future even when not showing the page controls

PR #121, opened 2023-03-28
[GLT-3857] Project Summary Filtering

hyeonjeongkwon:


ts/project-details.ts
```diff
@@ -66,7 +67,21 @@ function reload(definition: TableDefinition<any, any>) {
...
 }
+
+function handleSearch() {
+  const searchParams: Array<Pair<string, string>> = getSearchParams();
+  projectSummaryTable.addFilters(searchParams);
+  // for (const pair of searchParams) {
+  //   projectSummaryTable.applyFilters(pair.key, pair.value, true);
+  // }
+  return searchParams;
+}
```
djcooke, line 82: You can use `getSearchParams()` as the third parameter of both `TableBuilder`s to have the same effect. Then you don't need this function or `TableBuilder#addFilters`

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -404,6 +334,23 @@ public TableData<ProjectSummaryRow> getProjectSummaryRows(String name) {
     return data;
   }
 
+  public TableData<ProjectSummaryRow> getFilteredProjectSummaryRows(String projectName,
+      List<CaseFilter> filters) {
+    List<Case> cases = filterCases(caseData.getCases(), filters).toList();
+    Map<String, ProjectSummary> projectSummariesByName =
+        CaseLoader.calculateProjectSummaries(cases);
+    ProjectSummary projectSummary = projectSummariesByName.get(projectName);
```
djcooke, line 342: This works for filters that only apply at the case level, but is not working correctly for filters that should separate tests. For example: when using the "Test" filter, this returns numbers that include all tests for each case where one of the tests matches the filter. It should instead only count the matching test.

hyeonjeongkwon: Should we then only show project summary at the test level when "Test" filters are applied (ex. extraction, library prep., etc)?

djcooke: It should still count all gates. Example: say you have a case with 2 tests. Receipt is completed for the case, and extraction is completed for both tests.

* with no filter, this should show as having 2 completed receipt, and 2 completed extraction
* if you filter to match one of the tests, it should show 1 completed receipt and 1 completed extraction

This will make your filtering and calculation methods more complex. You'll need to filter the cases, keep that information, and also filter the tests, then calculate based only on those tests, but with the case-level data included too. Might want to create some structure like `List<Pair<Test, Case>>` or `Map<Case, <List<Test>>` when filtering

hyeonjeongkwon: What is the difference between creating the new structure and using the `filterTestTableViews`? Doesn't `TestTableViews` already includes case information?

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -713,22 +713,23 @@ private <T> List<T> loadFromJsonArrayFile(FileReader fileReader, ParseFunction<J
     return loaded;
   }
 
-  public static Map<String, ProjectSummary> calculateProjectSummaries(List<Case> cases) {
+  public static Map<String, ProjectSummary> calculateProjectSummaries(List<Case> caseList,
+      Map<Case, List<Test>> map) {
```
djcooke, line 717: One of these parameters will/should always be null, so a cleaner solution would be to make two separate methods. You could still share most of the code by separating all of the counting into an `addCounts(Case, List<Test>, Map<String, ProjectSummary.Builder>)` method. The bit at the end could also be moved into a buildProjectSummaries(Map<String, ProjectSummary.Builder>)` method to share

hyeonjeongkwon: @djcooke Do you think it would be better to add both methods to `caseloader`? Or one of them on `caseService`. Only `caseService` is using the filtered version, but I am wondering if it would be better style to put all of them together to minimize confusion in the future

djcooke: Both in `CaseLoader` is ideal so all of that logic is in one place

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -550,6 +497,22 @@ private Stream<TestTableView> filterTestTableViews(List<Case> cases,
     return stream;
   }
 
+  private Map<Case, List<Test>> getFilteredCaseAndTest(List<Case> cases,
+      Collection<CaseFilter> filters) {
+    Map<Case, List<Test>> testsByCase = filterCases(cases, filters)
+        .collect(Collectors.toMap(Function.identity(), x -> x.getTests()));
```
djcooke, line 503: No need for this to be a `Map` here. You can iterate over the `List` and use `Case::getTests` instead of `entry.getValue()` below

djcooke: You do need to create that map later. Currently, you are creating this map that contains all of the tests, and then you're replacing all of the values later. I'm suggesting you wait until later and create the map with the correct values initially

hyeonjeongkwon: Ah I see what you mean. Thank you

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -557,6 +504,22 @@ private Stream<TestTableView> filterTestTableViews(List<Case> cases,
...
+    Map<Case, List<Test>> testsByCase = new HashMap<>();
+    List<Case> filteredCases = filterCases(cases, filters).toList();
+    if (filters != null && !filters.isEmpty()) {
+      Map<CaseFilterKey, Predicate<Test>> filterMap =
+          buildFilterMap(filters, CaseFilter::testPredicate);
+      for (Predicate<Test> predicate : filterMap.values()) {
+        for (Case kase : filteredCases) {
+          testsByCase.put(kase, kase.getTests().stream().filter(predicate).toList());
+        }
+      }
```
djcooke, line 518: This is replacing the results of the previous predicate on each iteration. You'll need to filter the previous results if present. Something like:

```
List<Test> tests = testsByCase.getOrDefault(kase, kase.getTests());
testsByCase.put(kase, tests.stream().filter(predicate).toList());
```

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -557,6 +504,22 @@ private Stream<TestTableView> filterTestTableViews(List<Case> cases,
...
+    }
+    return testsByCase;
```
djcooke, line 520: This would return an empty map if called with no filters. Change the method to validate `filters` at the start since this should never happen. Throw NullPointerException if null, or IllegalArgumentException if empty

avarsava (approved):

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -506,14 +506,20 @@ private Stream<TestTableView> filterTestTableViews(List<Case> cases,
...
   private Map<Case, List<Test>> getFilteredCaseAndTest(List<Case> cases,
       Collection<CaseFilter> filters) {
+    if (filters == null) {
+      throw new NullPointerException("Filters cannot be null");
+    } else if (filters.isEmpty()) {
+      throw new IllegalStateException("Filters cannot be empty");
+    }
     Map<Case, List<Test>> testsByCase = new HashMap<>();
     List<Case> filteredCases = filterCases(cases, filters).toList();
     if (filters != null && !filters.isEmpty()) {
```
djcooke, line 516: Don't need this check anymore

djcooke: You don't need to check here because you were checking above. You removed the above checks instead of this one though

djcooke (approved):

PR #122, opened 2023-03-29
[GLT-3877] added Case QC Report page

djcooke:
Made a separate ticket to address the requisition QC group donor ID TODOs (https://jira.oicr.on.ca/browse/GLT-3888). Not important for clinical requisitions, which this is mainly intended for, but there could be some weirdness with RUO requisitions (of which there are still none in Dimsum anyway)

### DIFF
--- a/src/main/resources/templates/case-report.html
+++ b/src/main/resources/templates/case-report.html
@@ -18,22 +18,34 @@
     </div>
     <div id="caseAttributesContainer"></div>
 
-    <h2 class="text-green-200 font-sarabun font-light text-24 mt-4">Receipt/Inspection</h2>
+    <div class="keep-with-next">
+      <h2 class="text-green-200 font-sarabun font-light text-24 mt-4">Receipt/Inspection</h2>
+    </div>
     <div id="receiptTableContainer"></div>
 
-    <h2 class="text-green-200 font-sarabun font-light text-24 mt-4">Extraction</h2>
+    <div class="keep-with-next">
+      <h2 class="text-green-200 font-sarabun font-light text-24 mt-4">Extraction</h2>
+    </div>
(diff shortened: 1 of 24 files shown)


ts/case-report.ts
```diff
@@ -0,0 +1,637 @@
...
+    },
+  },
+  {
+    title: "Internal Name",
+    addContents(object, fragment) {
+      fragment.appendChild(
+        makeNameDiv(
+          object.donor.name,
+          urls.miso.sample(object.donor.id),
+          urls.dimsum.donor(object.donor.id)
```
hyeonjeongkwon, line 76: Shouldn't the url for dimsum use donor name not the id?

src/main/resources/templates/case-report.html
```diff
@@ -0,0 +1,43 @@
...
+
+    <h2 class="text-green-200 font-sarabun font-light text-24 mt-4">Library Preparation</h2>
+    <div id="libraryPreparationTableContainer"></div>
+
+    <h2 class="text-green-200 font-sarabun font-light text-24 mt-4">Library Qualification</h2>
+    <div id="libraryQualificationTableContainer"></div>
+
+    <h2 class="text-green-200 font-sarabun font-light text-24 mt-4">Full-Depth Sequencing</h2>
+    <div id="fullDepthSequencingTableContainer"></div>
+
```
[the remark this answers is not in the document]
djcooke: The case QC report (formerly Geneticist Review Report) is used before the clinical report is generated

ts/case-report.ts
djcooke: Yes. Good catch, thanks! Will fix

avarsava (approved):

hyeonjeongkwon (approved):

PR #125, opened 2023-04-12
[GLT-3876] Link Project Summary Counts to Filtered Views

hyeonjeongkwon:


ts/data/project-summary-row.ts
```diff
@@ -111,6 +112,14 @@ function displayCount(
     return;
   }
   fragment.appendChild(
-    document.createTextNode(projectSummaryField.count.toString())
+    // document.createTextNode(projectSummaryField.count.toString())
```
djcooke, line 115: remove commented code

ts/data/project.ts
```diff
@@ -139,12 +153,31 @@ export const projectDefinition: TableDefinition<ProjectSummary, void> = {
...
+
+const completedFilterKey = "COMPLETED";
+
+function displayFilteredProject(
+  count: number,
+  projectName: string,
+  filterValue: string
+) {
+  var url = new URL(document.location.href);
+  var params = url.searchParams;
```
djcooke, line 176: * You can construct a `new URLSearchParams()` without creating a `URL` first
* Use `const` instead of `var`

djcooke (commented):
This is applying the correct filters, but needs to also select the correct table. If you click one of the extraction counts, it should automatically select the Extractions table, for example. This will involve adding another query parameter to the Project Details page URL to specify a table to select by default

ts/data/project.ts
```diff
@@ -166,14 +166,17 @@ export const projectDefinition: TableDefinition<ProjectSummary, void> = {
...
 function displayFilteredProject(
   count: number,
   projectName: string,
   filterValue: string
 ) {
-  var url = new URL(document.location.href);
-  var params = url.searchParams;
+  const params = new URLSearchParams();
+  const tableFilterValue =
+    filterValue === "Informatics Review" ? filterValue : filterValue + "s";
```
djcooke, line 178: Please change the tab name to "Informatics Reviews" instead of making this exception. That way they're all consistently plural

djcooke (approved):

ts/data/project.ts
hyeonjeongkwon: Will do!

avarsava (approved):

PR #130, opened 2023-04-26
[GLT-3858] add date range filter

hyeonjeongkwon:


ts/component/table-builder.ts
```diff
@@ -15,7 +15,7 @@ import { showErrorDialog } from "./dialog";
 import { updateUrlParams } from "../util/urls";
 
 type SortType = "number" | "text" | "date";
-type FilterType = "text" | "dropdown";
+type FilterType = "text" | "dropdown" | "date";
```
djcooke, line 18: "text" can be used for the dates instead of adding this option. As is, it would cause an error if there were a date filter and the filters were not hidden

djcooke (commented):
It should not be possible to enter multiple date ranges. The values entered in the date pickers should be used, and choosing an option from the dropdown should set the dates in the date pickers. No additional filter element should be displayed

ts/data/project-summary-row.ts
```diff
@@ -26,7 +26,21 @@ export function getProjectSummaryRowDefinition(
...
   return {
     queryUrl: queryUrl,
-    filters: caseFilters,
+    filters: [
+      ...caseFilters,
+      {
+        title: "After Date",
+        key: "AFTER_DATE",
+        type: "date",
+        autocompleteUrl: urls.rest.autocomplete.testNames,
```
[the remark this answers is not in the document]
hyeonjeongkwon: Oops. I used it for testing forgot to remove it. I will remove it right now!

ts/component/table-builder.ts
hyeonjeongkwon: From what I understand, it would throw an error if filterType is "text" and has no autocomplete url. Should I remove the check? I thought it would be good to keep it for other filters

hyeonjeongkwon: Fair enough. Will do

src/main/resources/templates/project-detail.html
```diff
@@ -16,7 +16,28 @@ <h1 class="text-green-200 font-sarabun font-light font text-32" th:text="${title
...
+        <form action="" method="get" id="beforeDateForm"
+          class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex space-x-2 items-center">
+          <div>
+            <label for="date">Before Date:</label>
+            <input type="date" id="beforeDate" value="" />
+          </div>
+        </form>
+        <div id="projectSummaryTableFilter"></div>
+      </div>
+    </div>
```
djcooke, line 38: There seems to be a lot of unnecessary structure here. Form submission isn't being used and change listeners could be set on the date pickers instead, so the form elements aren't useful. There are a few divs that don't seem to have a purpose either. I believe this could be simplified down to something like

```
<div (flex)>
  <label>
  <date picker>
  <label>
  <date picker>
</div>
```

and then the dropdown can be appended to the existing div

ts/project-details.ts
```diff
@@ -74,8 +192,54 @@ function reload(definition: TableDefinition<any, any>) {
...
+      "afterDate"
+    ) as HTMLInputElement;
+    beforeDateCOntainer.value = filterValue;
+  } else {
+    const afterDateContainer = document.getElementById(
+      "beforeDate"
+    ) as HTMLInputElement;
+    afterDateContainer.value = filterValue;
+  }
+  reloadSummary();
```
djcooke, line 234: Can the table builder's `applyFilters` method be used instead of rebuilding the table entirely?

The pattern of calling `handleDateDropdownFilter` twice when choosing a range from the dropdown is causing two requests to happen, which isn't ideal. Maybe table builder's `applyFilters` should be renamed to `applyFilter`, and there should be another `applyFilters` method that accepts multiple filters to apply at once?

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -844,4 +854,47 @@ private static Map<String, ProjectSummary> buildProjectSummaries(
...
+    if (filters != null && !filters.isEmpty()) {
+      Map<DateFilterKey, Predicate<Sample>> filterMap = new HashMap<>();
+      for (DateFilter filter : filters) {
+        DateFilterKey key = filter.getKey();
+        if (filterMap.containsKey(key)) {
+          filterMap.put(key, filterMap.get(key).or(filter.samplePredicate()));
+        } else {
+          filterMap.put(key, filter.samplePredicate());
+        }
+      }
```
djcooke, line 869: It doesn't make sense to have multiple of the same date filter, so this mapping can be skipped. Each filter can be applied as is.

The results from this method are never really used - you just check if there are any matching items. You could simplify by renaming and changing this method to return a boolean instead:

```
samples.stream()
        .anyMatch(sample -> filters.stream().allMatch(filter -> filter.samplePredicate().test(sample)));
```

Can do similar for `filterRequisitionByQcGate`

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -794,26 +802,28 @@ private static void addCounts(Case kase, List<Test> tests,
...
       caseSummary.draftReportCompletedCount(testSize);
     }
     if (PendingState.DRAFT_REPORT.qualifyCase(kase) && !kase.isStopped()) {
       caseSummary.draftReportPendingCount(testSize);
     }
 
     // final report
-    if (kase.getRequisition().getFinalReports().stream()
-        .anyMatch(x -> x.isQcPassed())) {
+    if (CompletedGate.FINAL_REPORT.qualifyCase(kase)) {
```
djcooke, line 826: Date range not being applied to final reports

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -601,6 +614,9 @@ private ProjectSummaryRow getCompletedProjectSummaryRow(ProjectSummary projectSu
   }
 
   private ProjectSummaryRow getPendingProjectSummaryRow(ProjectSummary projectSummary) {
+    if (projectSummary == null) {
+      return new ProjectSummaryRow.Builder().title("Pending Work").build();
+    }
```
djcooke, line 619: Can be removed as `projectSummary` should never be null. Same in `getPendingQcProjectSummaryRow`

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/DateFilterKey.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.function.Function;
+import java.util.function.Predicate;
+import ca.on.oicr.gsi.dimsum.data.RequisitionQc;
+import ca.on.oicr.gsi.dimsum.data.Sample;
+
+
+public enum DateFilterKey {
+  // @formatter:off
+  AFTER_DATE(string -> {
+    return sample -> sample.getQcDate() != null &&!sample.getQcDate().isBefore(LocalDate.parse(string));
```
djcooke, line 13: For run-libraries (if sample.run is not null), data review date should be used instead of QC date. Same in `BEFORE_DATE`

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/DateFilterKey.java
```diff
@@ -0,0 +1,54 @@
...
+      }
+  };
+// @formatter:on
+  private final Function<String, Predicate<Sample>> create;
+
+  private DateFilterKey(Function<String, Predicate<Sample>> create) {
+    this.create = create;
+  }
+
+  public Function<String, Predicate<Sample>> create() {
```
djcooke, line 35: rename `samplePredicate()`

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/DateFilterKey.java
```diff
@@ -0,0 +1,54 @@
...
+  public String toString() {
+    switch (this) {
+      case AFTER_DATE:
+        return "AFTER_DATE";
+      case BEFORE_DATE:
+        return "BEFORE_DATE";
+      default:
+        return ""; // return empty string for unknown values
+    }
+  }
```
djcooke, line 53: Can remove since default `toString` already does this. Where you're calling this, it might be better to use `DateFilterKey#name` instead of `toString` also (though by default they return the same thing)

djcooke (approved):

PR #146, opened 2023-07-13
Add change note to changelog

callunity:
and remove change file from location where it didn't get picked up


djcooke (approved):
Oops! Thanks for catching that

avarsava (approved):

PR #150, opened 2023-08-04
[GLT-3764] Reduce run metric visual clutter

wuall826:
Collapse feature in run metric cells (Q30, Min Clusters/Reads (PF), and PhiX) to reduce visual clutter


ts/data/sample.ts
```diff
@@ -661,9 +661,32 @@ function addQ30Contents(
...
+    const isExpanded = contentWrapper.classList.toggle("hidden");
+    toggleButton.innerHTML = isExpanded ? "More" : "Less";
+  };
+
+  toggleButton.addEventListener("mouseover", () => {
+    toggleButton.style.textDecoration = "underline";
+  });
+  toggleButton.addEventListener("mouseout", () => {
+    toggleButton.style.textDecoration = "none";
+  });
```
djcooke, line 686: * Tailwind classes should be used for styling (colour, text decoration)
* Move button code to a separate function to reduce duplication

djcooke (commented):
* Can the more/less button be changed to an icon button beside the main line of text? Currently it takes up an entire line, and sometimes is only replacing a single line, defeating the purpose. It could use the same caret-down icon as on dropdowns and change to caret-up when expanded
* PhiX doesn't have a run-level value, so nothing is being displayed initially. I think the best option to make it consistent with others would be to show a calculated run-level value. This could be the minimum value from all lanes/reads, and display as the value + "+/R" e.g. "1.23+/R"
* The `addXContents` functions are also used by the Case QC Report, and the collapsing should not happen there - all values should be visible. You could add another boolean parameter to the functions to control whether collapsing happens

PR #151, opened 2023-08-09
[GLT-3938] use cardea

aryan2gandevia:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3938

✅ Includes a change file


src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -60,506 +39,118 @@ public class CaseLoader {
...
-   * @throws IOException if there is an error reading the timestamp file from disk
-   */
-  private ZonedDateTime getDataTimestamp() throws IOException {
-    String timeString = Files.readString(timestampFile.toPath()).trim();
-    if ("WORKING".equals(timeString)) {
-      return null;
+    try {
+      cardeaCaseData =
+          builder
+              .codecs(configurer -> configurer.defaultCodecs().maxInMemorySize(1024 * 1024 * 1024))
```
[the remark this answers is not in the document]
aryan2gandevia: Default is 256kb

src/main/java/ca/on/oicr/gsi/dimsum/data/ProjectSummary.java
```diff
@@ -2,8 +2,11 @@
 
 import static java.util.Objects.requireNonNull;
 import javax.annotation.concurrent.Immutable;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 
 @Immutable
+@JsonDeserialize(builder = ProjectSummary.Builder.class)
 public class ProjectSummary {
```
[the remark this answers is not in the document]
aryan2gandevia: There is no need, but just for consistency's sake and in case in the future for some reason it is needed, might as well add it there.

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
```diff
@@ -60,507 +38,108 @@ public class CaseLoader {
...
+    ZonedDateTime currentTimeStamp = builder.build().get().uri(cardeaUrl + "/timestamp").retrieve()
+        .bodyToMono(ZonedDateTime.class).block();
 
-      return caseData;
+    if (previousTimestamp != null && !currentTimeStamp.isAfter(previousTimestamp)) {
+      log.debug("Current case data is up to date with data files; aborting reload.");
+      return null;
     }
-  }
+    ca.on.oicr.gsi.cardea.data.CaseData cardeaCaseData = loadCardeaData(builder);
```
callunity, line 74: If Cardea returned an empty/null response for whatever reason, should Dimsum go ahead and replace its cache with the empty/null data?

aryan2gandevia: Maybe it should just retain the previous data as opposed to the null response, so I could treat it as a case where the logger has an error message describing what happened, and the function returns `null`?

callunity (commented):
Judging by the tests, Dimsum is expecting `Case` to have an `Assay` that it can query for values (especially for filtering : https://github.com/oicr-gsi/dimsum/blob/main/src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CaseFilterKey.java#L17 ) but Cardea's `Case` class only has the assayId and assayName, which is causing some tests to break.

src/main/java/ca/on/oicr/gsi/dimsum/CaseLoader.java
callunity: I think that's a good plan, and our existing monitoring should alert us for when this happens. You might need to update downstream functions that consume cardeaCaseData, so they don't run into NPEs.

src/test/resources/application-test.properties
```diff
@@ -1,6 +1,6 @@
 baseurl=http://localhost:8080
 instancename=
-datadirectory=${basedir}/src/test/resources/testdata
+cardea.url=https://cardea.gsi.oicr.on.ca
```
callunity, line 3: This should also probably be https://www.example.com

callunity (approved):
This looks good! And quite the massive change

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/TglTrackingReport.java
```diff
@@ -31,15 +30,13 @@ public class TglTrackingReport extends Report {
   private static final String METRIC_COVERAGE = "Mean Coverage Deduplicated";
   private static final String METRIC_CLUSTERS = "Pipeline Filtered Clusters";
 
-  private static Map<Long, Assay> assaysById = new HashMap<Long, Assay>();
+  public static class RowData {
```
djcooke, line 33: Could be private

djcooke (approved):

PR #152, opened 2023-08-16
[GLT-3764] Reduce run metric clutter

wuall826:
Collapse feature in run metric cells (Q30, Min Clusters/Reads (PF), and PhiX) to reduce visual clutter


ts/data/sample.ts
```diff
@@ -646,11 +654,57 @@ export function addMetricValueContents(
   }
 }
 
+function createCollapseButton(
+  contentWrapper: HTMLElement,
+  toggleText: string,
+  shouldCollapse: boolean
```
djcooke, line 660: This is only called when you want to collapse, and with an explicit value of `true`, so there's no need for the parameter

ts/data/sample.ts
```diff
@@ -858,31 +906,40 @@ function addPhixContents(
...
+    laneDiv.innerHTML = nullOrUndefined(lane.percentPfixRead1)
+      ? `${laneText}<span>${makeNotFoundIcon()}</span>`
+      : `${laneText}R1: ${lane.percentPfixRead1}${
+          !nullOrUndefined(lane.percentPfixRead2)
+            ? `; R2: ${lane.percentPfixRead2}`
+            : ""
+        }`;
+
+    if (addTooltip && !nullOrUndefined(lane.percentPfixRead1)) {
+      tooltip.addTarget(laneDiv, addContents);
```
djcooke, line 932: I don't see why there was any need to rewrite this, but it's now less readable, and also broken if any PhiX values are null:

image

ts/data/sample.ts
```diff
@@ -858,31 +906,40 @@ function addPhixContents(
...
     }
-    fragment.appendChild(laneDiv);
+
+    contentWrapper.appendChild(laneDiv);
   });
+
+  const minPhixDiv = document.createElement("div");
+  minPhixDiv.classList.add("whitespace-nowrap", "print-hanging");
```
djcooke, line 939: These classes should only be on the lane-level lines

ts/data/sample.ts
```diff
@@ -858,31 +912,47 @@ function addPhixContents(
...
         text += `; R2: ${lane.percentPfixRead2}`;
       }
-      laneDiv.innerText = text;
-      if (addTooltip) {
-        tooltip.addTarget(laneDiv, addContents);
-      }
     }
-    fragment.appendChild(laneDiv);
+
+    laneDiv.innerHTML = text;
```
djcooke, line 929: This is still going to result in the text "[object HTMLElement]" if values are missing because `makeNotFoundIcon` returns an object that is meant to be appended to the div as it originally was. `innerHtml` should not be used here, and is best avoided in most cases since it is more error-prone (e.g. here), and will perform poorer than creating and appending elements. Again, there was no reason to change this code which was previously working fine.

ts/data/sample.ts
```diff
@@ -858,8 +921,7 @@ function addPhixContents(
     metrics.forEach((metric) => addMetricRequirementText(metric, fragment));
   };
 
-  const multipleLanes = sample.run.lanes.length > 1;
-  sample.run.lanes.forEach((lane) => {
+  const processLane = (lane: Lane, multipleLanes: boolean) => {
```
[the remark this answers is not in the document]
djcooke: This was not resolved

djcooke (approved):

avarsava (approved):

PR #155, opened 2023-09-15
[GLT-3987] Removed QC report button from requisition details

wuall826:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3987

✅ Includes a change file


ts/details.ts
```diff
@@ -74,6 +74,11 @@ switch (tableContainer.getAttribute("data-detail-type")) {
     }
     const copyButton = makeCopyButton(requisitionName);
     titleMisoLink.parentNode?.insertBefore(copyButton, titleMisoLink);
+    const qcReportButton = document.getElementById("qcReportButton");
+    if (qcReportButton) {
+      qcReportButton.style.display = "none";
+    }
+    break;
```
djcooke, line 81: I think the `break` statement alone is the fix here. The button is created below, and without an ID, so I don't think the rest of this is doing anything?

Please add another `break` statement at the end of the `CASE_ID` case to prevent similar bugs if we expand this in the future.

changes/remove_qcReportButton
djcooke, line 1: This is a bug fix, so should be listed as a fix (start filename with `fix_`).

See `changes/README.md` for more info on creating change files

djcooke (commented):
This PR is deleting change files from other PRs?

wuall826:
> This PR is deleting change files from other PRs?

I think it's because I used the repository before the new dimsum release, so it contained the change file from other PRs commits.

wuall826:
Don't mind this commit I think something's wrong

callunity (approved):

djcooke (approved):

PR #156, opened 2023-09-21
[GLT-3991] add action buttons to the top

wuall826:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3991

✅ Includes a change file

### DIFF
--- a/changes/change_topButton
+++ b/changes/change_topButton
@@ -1,1 +1,1 @@
-Action buttons at both top and bottom of the table
+Action buttons now appear at both top and bottom of the table
--- a/changes/change_topButton
+++ b/changes/change_topButton
@@ -1,0 +1,1 @@
+Action buttons at both top and bottom of the table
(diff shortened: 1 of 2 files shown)


ts/component/table-builder.ts
```diff
@@ -226,6 +226,11 @@ export class TableBuilder<ParentType, ChildType> {
       this.addFilterControls(topControlsContainer);
       this.addPagingControls(topControlsContainer);
       this.container.appendChild(topControlsContainer);
+      if (this.definition.bulkActions || this.definition.staticActions) {
+        topControlsContainer.className =
+          "flex justify-end mt-4 items-top space-x-2";
+        this.addActionButtons(topControlsContainer);
+      }
```
djcooke, line 233: These should be displayed as similar to the ones on the bottom as possible

* include even when `disablePageControls` is `true`
* display left of the paging controls if present. The top paging controls include more than the bottom and these should not be split up, so it still won't look the exact same.

changes/change_topButton
```diff
@@ -0,0 +1 @@
+Action buttons at both top and bottom of the table
```
djcooke, line 1:
```suggestion
Action buttons now appear at both top and bottom of the table
```

djcooke (approved):

mlaszloffy (approved):

PR #158, opened 2023-10-12
[GLT-3979] replace default Spring error page with custom error page

wuall826:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3979

✅ Includes a change file

### DIFF
--- a/src/main/java/ca/on/oicr/gsi/dimsum/controller/MvcExceptionHandler.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/controller/MvcExceptionHandler.java
@@ -3,6 +3,7 @@ package ca.on.oicr.gsi.dimsum.controller;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.server.ResponseStatusException;
@@ -19,34 +20,33 @@ public class MvcExceptionHandler {
 
     @ExceptionHandler(ResponseStatusException.class)
     public ModelAndView handleResponseStatusException(ResponseStatusException ex) {
-        ModelAndView mav = prepareErrorModelAndView(ex.getStatus().value(), ex.getReason(), ex);
-
-        if (ex.getStatus().is5xxServerError()) {
-            logger.error("Error Status: " + ex.getStatus() + ". Error Message: " + ex.getReason(),
-                    ex);
-        } else {
-            logger.info("Error Status: " + ex.getStatus() + ". Error Message: " + ex.getReason());
-        }
-
-        return mav;
+        return prepareErrorModelAndView(ex.getStatus(), ex.getReason(), ex);
     }
 
     @ExceptionHandler(Exception.class)
     public ModelAndView handleGeneralException(Exception ex) {
-        logger.error("Unexpected error", ex);
-        return prepareErrorModelAndView(500, "Unexpected error", ex);
+        return prepareErrorModelAndView(HttpStatus.INTERNAL_SERVER_ERROR, "Unexpected error", ex);
     }
 
-    private ModelAndView prepareErrorModelAndView(int errorStatus, String errorMessage,
+    private ModelAndView prepareErrorModelAndView(HttpStatus status, String errorMessage,
             Exception ex) {
+        String resolvedErrorMessage =
+                (errorMessage != null) ? errorMessage : "No message available";
+        if (status.is5xxServerError()) {
+            logger.error("Error Status: {}. Error Message: {}", status, resolvedErrorMessage, ex);
+        } else {
+            logger.info("Error Status: {}. Error Message: {}", status, resolvedErrorMessage);
+        }
+
         ModelAndView mav = new ModelAndView("error");
-        mav.addObject("errorStatus",
-                errorStatus + (ex instanceof ResponseStatusException
-                        ? " " + ((ResponseStatusException) ex).getStatus().getReasonPhrase()
-                        : ""));
-        mav.addObject("errorMessage", errorMessage);
+        mav.addObject("errorStatus", status.value());
+        mav.addObject("errorReason", status.getReasonPhrase());
+        mav.addObject("errorMessage", resolvedErrorMessage);
         mav.addObject("dataAgeMinutes", commonModelAttributeProvider.getDataAgeMinutes());
         mav.addObject("buildVersion", commonModelAttributeProvider.getBuildVersion());
+
+        mav.addObject("title", status.value() + " " + status.getReasonPhrase() + " - Dimsum");
+
         return mav;
     }
 }
--- a/src/main/java/ca/on/oicr/gsi/dimsum/controller/MvcExceptionHandler.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/controller/MvcExceptionHandler.java
@@ -14,23 +14,19 @@ public class MvcExceptionHandler {
 
     @Autowired
     private CommonModelAttributeProvider commonModelAttributeProvider;
+
     private static final Logger logger = LoggerFactory.getLogger(MvcExceptionHandler.class);
 
     @ExceptionHandler(ResponseStatusException.class)
     public ModelAndView handleResponseStatusException(ResponseStatusException ex) {
-        ModelAndView mav = new ModelAndView("error");
-        mav.addObject("errorStatus", ex.getStatus().toString());
-        mav.addObject("errorMessage", ex.getReason());
-
-        mav.addObject("dataAgeMinutes", commonModelAttributeProvider.getDataAgeMinutes());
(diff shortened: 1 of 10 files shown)


callunity (approved):

src/main/java/ca/on/oicr/gsi/dimsum/controller/RestExceptionHandler.java
```diff
@@ -0,0 +1,38 @@
...
+public class RestExceptionHandler {
+
+    @ExceptionHandler(ResponseStatusException.class)
+    public ResponseEntity<Map<String, String>> handleResponseStatusException(
+            ResponseStatusException ex) {
+        Map<String, String> error = new HashMap<>();
+        error.put("status", ex.getStatus().toString());
+        error.put("error", ex.getReason());
+
+        return new ResponseEntity<>(error, ex.getStatus());
```
djcooke, line 22: REST exceptions should be logged similar to MVC exceptions. Again, ideal to just have another method that handles everything, which both of these handler methods can call

src/main/resources/templates/error.html
```diff
@@ -0,0 +1,31 @@
...
+  <div layout:fragment="content" class="m-4">
+
+    <h1 class="font-inter font-medium text-32 mb-4">
+      <span th:text="${errorStatus} ?: 'N/A'"></span>
+    </h1>
+
+    <div class="flex items-center mt-4 bg-pink border-2 border-red">
+      <i class="fas fa-exclamation-circle text-red ml-2"></i>
+      <p class="font-inter font-medium text-16 text-red ml-2" th:text="${errorMessage}"></p>
+    </div>
```
djcooke, line 20: This styling doesn't match the rest of Dimsum - coloured headings, rounded tables, etc.

What do you think about using the heading "Error" in red, then including the code, reason, and message in an `AttributeList` (like the top section of the Case QC Report)?

src/main/resources/templates/error.html
```diff
@@ -0,0 +1,31 @@
...
+
+    <div class="mt-4">
+      <a href="/" class="text-green-200 hover:underline font-bold block mb-2">Go to Home</a>
```
djcooke, line 23: This is not necessary since the logo at the top left is always present to link to the home page

src/main/java/ca/on/oicr/gsi/dimsum/controller/MvcExceptionHandler.java
```diff
@@ -14,32 +14,39 @@ public class MvcExceptionHandler {
...
-        mav.addObject("buildVersion", commonModelAttributeProvider.getBuildVersion());
+        ModelAndView mav = prepareErrorModelAndView(ex.getStatus().value(), ex.getReason(), ex);
 
         if (ex.getStatus().is5xxServerError()) {
-            logger.error(ex.getReason(), ex);
+            logger.error("Error Status: " + ex.getStatus() + ". Error Message: " + ex.getReason(),
+                    ex);
+        } else {
+            logger.info("Error Status: " + ex.getStatus() + ". Error Message: " + ex.getReason());
         }
```
djcooke, line 29: Move logging into `prepareErrorModelAndView` - it should be handled the same for both

src/main/java/ca/on/oicr/gsi/dimsum/controller/MvcExceptionHandler.java
```diff
@@ -14,32 +14,39 @@ public class MvcExceptionHandler {
...
         return mav;
     }
 
     @ExceptionHandler(Exception.class)
     public ModelAndView handleGeneralException(Exception ex) {
         logger.error("Unexpected error", ex);
+        return prepareErrorModelAndView(500, "Unexpected error", ex);
+    }
+
+    private ModelAndView prepareErrorModelAndView(int errorStatus, String errorMessage,
```
djcooke, line 40: Take an `HttpStatus` instead of `int`. `handleGeneralException` can pass `HttpStatus.INTERNAL_SERVER_ERROR` (or whatever it's called). Then you always have both the code and reason phrase, and you don't need to try to pull it from the exception below

src/main/java/ca/on/oicr/gsi/dimsum/controller/RestExceptionHandler.java
```diff
@@ -2,37 +2,47 @@
...
+    }
 
-        return new ResponseEntity<>(error, HttpStatus.INTERNAL_SERVER_ERROR);
+    private void logException(int status, String message, Exception ex) {
+        if (status >= 500) {
+            logger.error("Error Status: " + status + ". Error Message: " + message, ex);
+        } else {
+            logger.info("Error Status: " + status + ". Error Message: " + message);
+        }
     }
```
djcooke, line 39: Make this part of `prepareErrorResponse` or call it from there, so each handler method doesn't need to call both methods

tailwind.config.js
```diff
@@ -24,7 +24,6 @@ module.exports = {
       black: "#28292F",
       yellow: "#F4F5AD",
       red: "#DF8484",
-      pink: "#FFF6F6",
```
djcooke: Need to regenerate the css to use this change

src/main/java/ca/on/oicr/gsi/dimsum/controller/MvcExceptionHandler.java
```diff
@@ -14,32 +14,39 @@ public class MvcExceptionHandler {
...
+            Exception ex) {
         ModelAndView mav = new ModelAndView("error");
-        mav.addObject("errorMessage", "Unexpected error");
+        mav.addObject("errorStatus",
+                errorStatus + (ex instanceof ResponseStatusException
+                        ? " " + ((ResponseStatusException) ex).getStatus().getReasonPhrase()
+                        : ""));
```
djcooke, line 46: Add the code and reason phrase as separate properties and display in separate rows on front-end titled "Code" and "Error."

It would also be good to add these both to the title, so it ends up as something like "500 Internal Server Error - Dimsum"

src/main/resources/templates/error.html
```diff
@@ -10,17 +10,29 @@
...
+          <td class="p-4 border-grey-200 border-l-1 text-left align-text-top" th:text="${errorStatus} ?: 'N/A'"></td>
+        </tr>
+        <tr>
+          <th
+            class="p-4 text-white font-semibold bg-grey-300 text-left align-text-top whitespace-nowrap border-grey-200 border-t-1">
+            Message</th>
+          <td class="p-4 border-grey-200 border-l-1 text-left align-text-top border-t-1" th:text="${errorMessage}"></td>
+        </tr>
+      </tbody>
+    </table>
```
djcooke, line 33: Make use of `attribute-list.ts` to generate this instead. You can see `case-report.ts` for an example of that. Some notes:

* Add the model attributes as data properties on the div container for the attribute list. Then you can read these into a typescript object
* You'll need to add a new script for the error page, which must be configured as an entry point in `webpack.config.js`. Create the attribute definitions and then build the attribute list on page load

This solution is a bit more complex, but keeps the project more maintainable as all attribute lists are handled in the same way by the same code.

djcooke (commented):
There is a class of errors that this is partially handling - the error page is displayed, but none of the model attributes are there. For an example, try navigating to `http://localhost:<port>/asdf`

ts/error.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+  },
+];
+
+document.addEventListener("DOMContentLoaded", () => {
+  const container = document.getElementById("error-attribute-list");
+
+  if (container) {
+    const errorData: ErrorData = {
+      status: container.dataset.status ?? "N/A",
```
djcooke, line 35: "Unknown" would make more sense than "N/A"
Can use `||` instead of `??` to take care of empty string too. Same for `reason`

ts/error.ts
```diff
@@ -0,0 +1,53 @@
...
+      reason: container.dataset.reason ?? "Unknown Reason",
+      message: container.dataset.message ?? "",
```
djcooke, line 37: Can make message optional in `ErrorData` instead of setting it to empty string

src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/CustomErrorController.java
```diff
@@ -0,0 +1,32 @@
...
+        ModelAndView modelAndView = new ModelAndView();
+        modelAndView.setViewName("error");
+
+        int statusCode = status != null ? Integer.parseInt(status.toString()) : 500;
+        String statusMessage = HttpStatus.valueOf(statusCode).getReasonPhrase();
+
+        modelAndView.addObject("errorStatus", statusCode);
+        modelAndView.addObject("errorReason", statusMessage);
+
+        modelAndView.addObject("title", statusCode + " " + statusMessage + " - Dimsum");
```
djcooke, line 27: Title is already set in the html template, so this isn't needed

djcooke (commented):
Looks great! Just a few minor fixups

src/main/resources/application.yml
```diff
@@ -28,3 +28,7 @@ spring:
 
 build:
   version: "@project.version@"
+
+server:
+  error:
+    path:/error
```
djcooke, line 34: Missing space after `:`. This is causing the test failure

PR #163, opened 2023-11-10
[GLT-4008] added incomplete filter to case QC report page

wuall826:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4008

- [ ] Includes a change file

### DIFF
--- a/changes/add_GLT-4008.md
+++ b/changes/add_GLT-4008.md
@@ -1,0 +1,1 @@
+Incomplete filter to Case QC Report page
--- a/src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CaseFilterKey.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CaseFilterKey.java
@@ -72,6 +72,25 @@ public enum CaseFilterKey {
       return string -> getGate(string).requisitionPredicate();
     }
   },
+  INCOMPLETE(string -> {
+    CompletedGate gate = getGate(string);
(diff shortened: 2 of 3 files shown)


grapigeau (approved):

djcooke (commented):
This is a nice, clean solution!

I think there is a problem with the completed filter that affects this though. For test-level gates (extraction, library prep, library qual, full depth), the completed filter includes items where _any_ test has completed the gate. If a case has multiple tests and only some are completed the gate, they should be considered incomplete. Can you fix that?

Also please add tests for the incomplete filter, including one to test a case where the gate has some tests completed and others incomplete (case should be considered incomplete)

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CompletedGate.java
```diff
@@ -198,7 +178,7 @@ public Predicate<Sample> samplePredicate(MetricCategory requestCategory) {
   }
 
   public boolean qualifyCase(Case kase) {
-    return kase.getTests().stream().allMatch(test -> qualifyTest(test));
+    return kase.getTests().stream().anyMatch(test -> qualifyTest(test));
```
djcooke, line 181: This is changing it back to how it was. It should be left as `allMatch` so that a gate is only considered complete for the case if it is complete for all of the tests

grapigeau (approved):

djcooke (approved):

PR #168, opened 2023-12-07
[GLT-4034] added paused cases handling

wuall826:
Jira ticket or GitHub issue: 
https://jira.oicr.on.ca/browse/GLT-4034

- [x] Includes a change file

### DIFF
--- a/ts/component/table-components.ts
+++ b/ts/component/table-components.ts
@@ -2,8 +2,7 @@ import { siteConfig } from "../util/site-config";
 import { urls } from "../util/urls";
 import { FilterDefinition, Sort } from "./table-builder";
 
-const stopStatus: string[] = ["Yes", "No"];
-const pauseStatus: string[] = ["Yes", "No"];
+const yesNoOptions: string[] = ["Yes", "No"];
 
 export const latestActivitySort: Sort = {
   columnTitle: "Latest Activity",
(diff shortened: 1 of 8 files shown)


ts/data/case.ts
```diff
@@ -577,6 +592,8 @@ export function handleNaSamplePhase(
   if (requisition.stopped && !samples.length) {
     addNaText(fragment);
     return true;
+  } else if (requisition.paused) {
```
grapigeau, line 595: Is there a reason to specify this separately?

ts/component/table-components.ts
```diff
@@ -3,6 +3,7 @@ import { urls } from "../util/urls";
 import { FilterDefinition, Sort } from "./table-builder";
 
 const stopStatus: string[] = ["Yes", "No"];
+const pauseStatus: string[] = ["Yes", "No"];
```
djcooke, line 6: Could rename to something like `yesNoOptions` and use the same array for both

ts/util/html-utils.ts
```diff
@@ -68,14 +68,15 @@ export function addMisoIcon(container: HTMLElement, url: string) {
...
-export type CellStatus = "na" | "stopped" | "warning" | "error";
+export type CellStatus = "na" | "stopped" | "warning" | "error" | "paused";
 
 // map cell status to styles
 const highlightClasses: Record<CellStatus, string[]> = {
   na: ["text-grey-300", "bg-grey-150"],
   stopped: ["bg-grey-100"],
   warning: ["bg-yellow"],
   error: ["bg-red"],
+  paused: ["bg-grey-100"],
```
djcooke, line 79: Could use a single status for stopped and paused since the style is the same. Maybe rename to "disabled" for clarity

grapigeau (approved):

djcooke (commented):
Incomplete steps for paused cases still look pending in the case list:

image

* There should be no yellow highlights. You can update `getSamplePhaseHighlight` and `getRequisitionPhaseHighlight` to return null if paused
* There should be no pending work icon (construction icon) when the case is paused, which you'll have to update in each of the column definitions

djcooke (approved):

PR #174, opened 2024-01-09
[GLT-4046] added QC status sort in qc step tables

wuall826:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4046

- [x] Includes a change file


src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/QCStatusSort.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+import ca.on.oicr.gsi.cardea.data.Sample;
+
+public class QCStatusSort implements Comparator<Sample> {
+
+    private boolean reverseOrder;
+
+    public void setReverseOrder(boolean reverseOrder) {
+        this.reverseOrder = reverseOrder;
+    }
```
djcooke, line 13: There's no need to handle reversing - The `Comparator` interface has a default methods to do this

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/QCStatusSort.java
```diff
@@ -0,0 +1,46 @@
...
+        } else if (PendingState.Helpers.isPendingDataReview(sample)) {
+            return "dataReview";
+        } else if (PendingState.Helpers.isTopUpRequired(sample)) {
+            return "topUp";
+        } else if (PendingState.Helpers.isPassed(sample)) {
+            return "passed";
+        } else {
+            return "failed";
+        }
+    }
```
djcooke, line 30: Can simplify this by returning integers instead. Then there's no need for the Strings, the order collection, or lookup in your `compare` method - just compare the values returned from this method using `Integer.compare`

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/QCStatusSort.java
```diff
@@ -0,0 +1,46 @@
+package ca.on.oicr.gsi.dimsum.service.filtering;
+
+import java.util.Comparator;
+import java.util.List;
+import ca.on.oicr.gsi.cardea.data.Sample;
+
+public class QCStatusSort implements Comparator<Sample> {
```
djcooke, line 7: This can be included in the `SampleSort` enum declaration since it's not used elsewhere. See `CaseSort.URGENCY` for an example of what this looks like

src/test/java/ca/on/oicr/gsi/dimsum/service/filtering/QCStatusSortTest.java
```diff
@@ -0,0 +1,26 @@
...
+    qcStatusSort.setReverseOrder(false);
+    Collections.sort(QC_STATUS_ORDER, Collections.reverseOrder());
+  }
+
+  @Test
+  public void testSortByStatusDescending() {
+    QCStatusSort qcStatusSort = new QCStatusSort();
+    qcStatusSort.setReverseOrder(true);
+    Collections.sort(QC_STATUS_ORDER);
+  }
```
djcooke, line 25: * These tests are not actually using `QCStatusSort` to sort `QC_STATUS_ORDER`, and they couldn't because `QCStatusSort` compares Samples and `QC_STATUS_ORDER` contains Strings. `Collections#sort` and `Collections.reverseOrder` both use natural order - not `QCStatusSort`
* The tests are not making any assertions to check whether the correct results were achieved
* Instead of this separate test class, please add tests to `SampleSortTest` that test sorting Samples using `SampleSort.QC_STATUS`

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/PendingState.java
```diff
@@ -338,7 +338,7 @@ protected Function<Case, Stream<Sample>> getFilteredSamples(MetricCategory reque
     throw new IllegalStateException("This gate does not apply to samples");
   }
 
-  private static class Helpers {
+  public static class Helpers {
```
djcooke, line 341: Please create a public `SampleUtils` class in the `util` package and move the necessary methods to that instead of making parts of this inner class public. This way, they're easier to find and reference

src/test/java/ca/on/oicr/gsi/dimsum/service/filtering/SampleSortTest.java
```diff
@@ -28,29 +31,81 @@ public void testSortByNameDescending() {
     assertOrder(samples, Sample::getName, namesOrdered, true);
   }
 
+  @Test
+  public void testSortByQcStatusAscending() {
+    List<Sample> samples = getSamplesSorted(SampleSort.QC_STATUS, false);
+    assertOrder(samples, SampleSort::getSampleQcStatus, qcStatusOrdered, false);
```
djcooke, line 37: This is testing that the statuses are ordered, but not testing that the correct status ints are being returned for each sample. Maybe leave this as is, but also add tests for `SampleSort#getSampleQcStatus`. They should look something like:

* mock sample
* mock all relevant attributes, similar to what you did in `mockSample` below
* assert that `getSampleSort` returns the expected value for that sample

Include one for every possible status int.

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/SampleSort.java
```diff
@@ -37,4 +39,17 @@ public Comparator<Sample> comparator() {
     return comparator;
   }
 
+  public static int getSampleQcStatus(Sample sample) {
```
djcooke, line 42: Change to protected. Getting an int to represent QC status doesn't really make sense outside of this class, so ideally would be private, but protected makes it visible for testing

djcooke (commented):
I think you reversed some of your changes?

djcooke (approved):

changes/add_GLT-4046
```diff
@@ -0,0 +1 @@
+Sort by QC status in all QC step tables
```
djcooke, line 1: The sort was not added to all QC step tables - it's not on analysis review, release approval, or release. Please update the change message to specify receipt, extraction, library qualification, and full-depth sequencing tables.

grapigeau (approved):

PR #177, opened 2024-01-25
[GLT-4084] fixed errors in QC in MISO function

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4084

- [x] Includes a change file

### DIFF
--- a/ts/data/sample.ts
+++ b/ts/data/sample.ts
@@ -24,7 +24,6 @@ import {
   getDivisorUnit,
   getMetricNames,
   getMetricRequirementText,
-  getSingleThreshold,
   makeMetricDisplay,
   makeNotFoundIcon,
   makeStatusIcon,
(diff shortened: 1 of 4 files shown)


ts/data/sample.ts
```diff
@@ -463,15 +463,42 @@ function getSampleMetrics(
...
-      return {
+      const misoMetric: MisoRunLibraryMetric = {
         title: metric.name,
         threshold_type: metric.thresholdType.toLowerCase(),
         threshold: getSingleThreshold(metric),
         value: revisedValue,
       };
+      if (metric.thresholdType == "BETWEEN") {
+        if (metric.maximum === undefined) {
+          throw new Error("Metric is missing maximum value");
```
avarsava, line 474: why do we only test for maximum here?

avarsava (approved):

ts/data/sample.ts
djcooke: The minimum is checked when populating it in `threshold` above

wuall826 (approved):

PR #184, opened 2024-02-27
enforce node and npm versions

djcooke:
Jira ticket or GitHub issue: n/a

The `lockFileVersion` keeps changing with people using different versions of npm, and this will prevent that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,11 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
-          distribution: 'temurin'
-          java-version: '17'
-          cache: 'maven'
+          distribution: "temurin"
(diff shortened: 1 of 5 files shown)


wuall826 (approved):

.github/workflows/ci.yml
```diff
@@ -13,11 +13,11 @@ jobs:
...
-          java-version: '17'
-          cache: 'maven'
+          distribution: "temurin"
+          java-version: "17"
+          cache: "maven"
       - uses: actions/setup-node@v3
         with:
-          node-version: 17
-          cache: 'npm'
+          node-version: 20
```
grapigeau, line 21: Need to update the readme with the new required version

djcooke: Updated. Thanks for catching that

grapigeau (approved):

PR #203, opened 2024-04-24
[GLT-4069] added case start and end date filter

wuall826:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4069

- [x] Includes a change file
- [x] Updated tests (or n/a)

### DIFF
--- a/changes/add_GLT-4069
+++ b/changes/add_GLT-4069
@@ -1,0 +1,1 @@
+case start and end date filter
--- a/changes/add_GLT-4069
+++ b/changes/add_GLT-4069
@@ -1,1 +1,1 @@
-case start and end date filter
+case start and end date filters
--- a/src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/MvcUtils.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/MvcUtils.java
@@ -19,10 +19,6 @@ public class MvcUtils {
 
   private static final String BEFORE_DATE_KEY = "BEFORE_DATE";
   private static final String AFTER_DATE_KEY = "AFTER_DATE";
-  private static final String STARTED_BEFORE_KEY = "STARTED_BEFORE";
-  private static final String STARTED_AFTER_KEY = "STARTED_AFTER";
-  private static final String COMPLETED_BEFORE_KEY = "COMPLETED_BEFORE";
-  private static final String COMPLETED_AFTER_KEY = "COMPLETED_AFTER";
 
   public static void validateDataQuery(DataQuery query) {
     if (query.getPageNumber() < 1) {
@@ -130,22 +126,6 @@ public class MvcUtils {
         .map(KeyValuePair::getValue)
         .findFirst().orElse(null);
     return value == null ? null : LocalDate.parse(value);
-  }
-
-  public static LocalDate parseStartedBeforeDate(DataQuery query) {
-    return parseCaseDateFilterValue(query, STARTED_BEFORE_KEY);
-  }
-
-  public static LocalDate parseStartedAfterDate(DataQuery query) {
-    return parseCaseDateFilterValue(query, STARTED_AFTER_KEY);
-  }
-
-  public static LocalDate parseCompletedBeforeDate(DataQuery query) {
-    return parseCaseDateFilterValue(query, COMPLETED_BEFORE_KEY);
-  }
-
-  public static LocalDate parseCompletedAfterDate(DataQuery query) {
-    return parseCaseDateFilterValue(query, COMPLETED_AFTER_KEY);
   }
 
   public static LocalDate parseAfterDate(DataQuery query) {
--- a/src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/MvcUtils.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/MvcUtils.java
@@ -19,7 +19,10 @@ public class MvcUtils {
 
   private static final String BEFORE_DATE_KEY = "BEFORE_DATE";
   private static final String AFTER_DATE_KEY = "AFTER_DATE";
-
+  private static final String STARTED_BEFORE_KEY = "STARTED_BEFORE";
+  private static final String STARTED_AFTER_KEY = "STARTED_AFTER";
+  private static final String COMPLETED_BEFORE_KEY = "COMPLETED_BEFORE";
+  private static final String COMPLETED_AFTER_KEY = "COMPLETED_AFTER";
 
   public static void validateDataQuery(DataQuery query) {
     if (query.getPageNumber() < 1) {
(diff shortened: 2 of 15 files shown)


src/test/java/ca/on/oicr/gsi/dimsum/MockCase.java
```diff
@@ -412,16 +416,43 @@ private static Case makeCase32() {
...
+
+    addCompletionDate(kase, completionDate);
+
+    return kase;
+  }
+
   private static Case makeCase(String donorName, String assayName, String projectName,
-      String requisitionName, int caseNumber) {
+      String requisitionName, int caseNumber, LocalDate startDate,
+      LocalDate completionDate) {
```
djcooke, line 455: These should be made to function like normal cases in any way that matters to the tests. Start date should never be null

src/test/java/ca/on/oicr/gsi/dimsum/MockCase.java
```diff
@@ -438,9 +469,35 @@ private static Case makeCase(String donorName, String assayName, String projectN
     addRequisition(kase, caseNumber, requisitionName);
     when(kase.getDeliverables()).thenReturn(new ArrayList<>());
     addDeliverable(kase, DeliverableType.DATA_RELEASE, "Full Pipeline");
+    when(kase.getStartDate()).thenReturn(startDate);
+    addCompletionDate(kase, completionDate);
     return kase;
   }
 
+  private static LocalDate addCompletionDate(Case kase, LocalDate completionDate) {
```
djcooke, line 477: As is, this method doesn't always do what its name implies, and a lot of release QCs are left without completion dates. Instead, set the date when marking release QC (add date parameter to `markRelease`)

ts/data/case.ts
```diff
@@ -184,7 +184,29 @@ export const caseDefinition: TableDefinition<Case, Test> = {
...
+    {
+      title: "Completed After",
+      key: "COMPLETED_AFTER",
+      type: "text",
+    },
+    {
+      title: "Completed Before",
+      key: "COMPLETED_BEFORE",
+      type: "text",
+    },
```
djcooke, line 208: These should be fine to add to the shared `caseFilters` constant. Otherwise they should work on all of the tables on the detail pages, but they'll only appear as options on the Cases tab

ts/component/table-builder.ts
```diff
@@ -251,6 +263,7 @@ export class TableBuilder<ParentType, ChildType> {
     if (!this.definition.disablePageControls) {
       this.addSortControls(topControlsContainer);
       this.addFilterControls(topControlsContainer);
+      this.addDateFilterControls(topControlsContainer);
```
djcooke, line 266: These should not be separate controls. They should be handled the same as other filters for the table, except with a new `FilterType` to handle date input instead of text

grapigeau (approved):

src/main/java/ca/on/oicr/gsi/dimsum/controller/mvc/MvcUtils.java
```diff
@@ -118,6 +117,14 @@ private static OmittedSampleFilter parseOmittedSampleFilter(KeyValuePair pair) {
...
   }
 
+  public static LocalDate parseCaseDateFilterValue(DataQuery query, String key) {
+    String value = query.getFilters().stream()
+        .filter(x -> key.equals(x.getKey()))
+        .map(KeyValuePair::getValue)
+        .findFirst().orElse(null);
+    return value == null ? null : LocalDate.parse(value);
+  }
+
```
djcooke, line 127: Unused

changes/add_GLT-4069
```diff
@@ -0,0 +1 @@
+case start and end date filter
```
djcooke, line 1:
```suggestion
case start and end date filters
```

src/main/java/ca/on/oicr/gsi/dimsum/service/filtering/CaseFilterKey.java
```diff
@@ -137,4 +152,24 @@ private static CompletedGate getGate(String label) {
...
+    }
+    if (releases.stream()
+        .anyMatch(release -> release.getQcPassed() == null || !release.getQcPassed())) {
+      return null;
+    }
+    return releases.stream()
+        .map(CaseRelease::getQcDate)
+        .max(LocalDate::compareTo)
+        .orElse(null);
+  }
```
djcooke, line 174: The exact same method exists in `CaseTatReport`. Move this to `SampleUtils` and use for both (In the report, it takes `RowData` as a parameter, but this can easily bechanged to a `Case` instead like here.

Maybe rename `SampleUtils` to `DataUtils` so it makes more sense to include utility methods for samples, cases, or any other data

ts/component/date-input.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.container.appendChild(this.dateInput);
+
+    // remove date input on 'Esc' key press
+    document.addEventListener("keydown", (event) => {
+      if (event.key === "Esc" || event.key === "Escape") {
+        this.container.remove();
+      }
+    });
+    // remove date input when mousedown outside of the date input
+    document.addEventListener("mousedown", (event) => {
```
djcooke, line 32: These event listeners are never removed and new ones are added to the document each time a `DateInput` is created. You can create a `destroy` function that removes the element and all listeners, and call that everywhere that you're currently calling `this.container.remove`. The listeners will have to be assigned to variables so you can reference them to remove

src/main/java/ca/on/oicr/gsi/dimsum/util/DataUtils.java
```diff
@@ -49,4 +54,29 @@ public static boolean isTopUpRequired(Sample sample) {
   public static boolean isAnalysisReviewSkipped(Case kase) {
     return kase.getProjects().stream().allMatch(project -> project.isAnalysisReviewSkipped());
   }
+
+  public static LocalDate getCompletionDate(Case kase, boolean returnStopped) {
+    if (returnStopped && kase.isStopped()) {
+      return null;
+    }
```
djcooke, line 61: This part should be removed

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/CaseTatReport.java
```diff
@@ -107,7 +107,11 @@ public Test getTest() {
                     .map(date -> date.format(DateTimeFormatter.ISO_LOCAL_DATE)).orElse(null);
               }),
               Column.forInteger("Release Days", x -> x.getCase().getReleaseDaysSpent()),
-              Column.forString("Completion Date", x -> getCompletionDate(x.getCase())),
+              Column.forString("Completion Date", x -> {
+                LocalDate date = DataUtils.getCompletionDate(x.getCase(), true);
+                return date != null ? date.format(DateTimeFormatter.ISO_LOCAL_DATE)
+                    : x.getCase().isStopped() ? "STOPPED" : null;
+              }),
```
djcooke, line 114: This report has been updated upstream. There's now a separate column to indicate whether the case is stopped, and this should show the completion date even if stopped

ts/component/date-input.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.container = document.createElement("div");
+    this.container.className =
+      "font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 inline-flex";
+
+    this.dateInput = document.createElement("input");
+    this.dateInput.type = "date";
+    this.dateInput.onchange = () => {
+      if (this.dateInput.value) {
+        onChange(this.dateInput.value);
+        this.container.remove(); // remove the date input after selection
```
djcooke, line 17: use `this.destroy()` instead

djcooke (approved):

PR #219, opened 2024-06-27
[GLT-4070]added TAT trend report

wuall826:
Jira ticket or GitHub issue:

- [x] Includes a change file
- [ ] Updated tests (or n/a)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "webpack-cli": "^4.9.2"
   },
   "dependencies": {
-    "plotly.js-dist-min": "^2.33.0",
-    "xlsx": "^0.18.6"
+    "plotly.js-dist-min": "^2.33.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -13,10 +13,15 @@
   },
   "license": "MIT",
   "devDependencies": {
+    "@types/plotly.js-dist-min": "^2.3.4",
     "tailwindcss": "^3.0.24",
     "ts-loader": "^9.3.0",
     "typescript": "^5.3.3",
     "webpack": "^5.76.0",
     "webpack-cli": "^4.9.2"
+  },
+  "dependencies": {
+    "plotly.js-dist-min": "^2.33.0",
+    "xlsx": "^0.18.6"
   }
 }
--- a/src/main/resources/templates/tat-trend.html
+++ b/src/main/resources/templates/tat-trend.html
@@ -50,7 +50,7 @@
           <label class="font-inter font-medium text-12"><input type="checkbox" value="Release Approval" checked> Release
             Approval</label>
           <label class="font-inter font-medium text-12"><input type="checkbox" value="Release" checked> Release</label>
-          <label class="font-inter font-medium text-12"><input type="checkbox" value="All Completed" checked> All
+          <label class="font-inter font-medium text-12"><input type="checkbox" value="Full Case" checked> All
             Completed</label>
         </div>
       </div>
--- a/src/main/resources/templates/tat-trend.html
+++ b/src/main/resources/templates/tat-trend.html
@@ -1,0 +1,62 @@
+<!DOCTYPE html>
+<html lang="en" xmlns:th="http://www.thymeleaf.org" xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout"
+  layout:decorate="~{layout}">
+
+<head>
+  <meta charset="UTF-8" />
+  <title>TAT Trend Report</title>
+</head>
+
+<body>
+  <div layout:fragment="content">
+    <h1 class="text-green-200 font-sarabun font-light text-32">TAT Trend Report</h1>
+    <div id="trendReportContainer">
+      <div id="controlsContainer">
+        <div id="groupingButtons" style="display: flex; gap: 10px; margin-top: 10px;">
+          <button id="weekButton"
+            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex space-x-2 items-center">Week</button>
+          <button id="monthButton"
+            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex space-x-2 items-center">Month</button>
+          <button id="quarterButton"
+            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex space-x-2 items-center">Fiscal
+            Quarter</button>
+          <button id="yearButton"
+            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex space-x-2 items-center">Fiscal
+            Year</button>
+          <div style="margin-left: auto; display: flex; align-items: center; gap: 10px;">
+            <label class="font-inter font-medium text-12">
+              <input type="checkbox" id="toggleAnnotations" checked> Show Annotations
+            </label>
+            <select id="dataSelection"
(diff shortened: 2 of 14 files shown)


wuall826:
the show annotation checkbox is there but the script for annotation still needs adjustment, so just ignore it.

ts/tat-trend.ts
```diff
@@ -0,0 +1,365 @@
...
+          selectedDataType
+        );
+
+        if ((completedDate && days != null) || gate === "All Completed") {
+          const date = new Date(
+            gate === "All Completed"
+              ? releaseCompletedDate || new Date()
+              : completedDate
+          );
+          const groupKey = getGroupKey(date, selectedGrouping);
```
djcooke, line 150: This is grouping by the completion date of the individual QC gate/step, and I'm not sure that's what we want. The other option is to group all steps by case completion date, which is what I am thinking we want, but I'm not sure.

The advantage to always using the case completion date is that if you filter by case completion date (e.g. cases completed in 2023), then all data will be in that range. With this implementation, you can filter to include cases completed in 2023 and some of your data will be outside of that - cases started in 2022 and finished in 2023 for example.

The current implementation includes cases that are incomplete, and I'm not sure whether we want that either.

Should discuss this more with QA

package.json
```diff
@@ -13,10 +13,15 @@
...
+    "@types/plotly.js-dist-min": "^2.3.4",
     "tailwindcss": "^3.0.24",
     "ts-loader": "^9.3.0",
     "typescript": "^5.3.3",
     "webpack": "^5.76.0",
     "webpack-cli": "^4.9.2"
+  },
+  "dependencies": {
+    "plotly.js-dist-min": "^2.33.0",
+    "xlsx": "^0.18.6"
```
djcooke, line 25: Getting this error when I try to build:

```
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/xlsx/-/xlsx-0.18.6.tgz - Not found
npm ERR! 404 
npm ERR! 404  'xlsx@https://registry.npmjs.org/xlsx/-/xlsx-0.18.6.tgz' is not in this registry.
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

I think it would be ideal to remove this library anyway though

ts/tat-trend.ts
wuall826: going with  group all steps by case completion date for now, will check with QA

avarsava (commented):
- Why do some data points disappear when i uncheck 'all completed' but leave all the other gates checked?
- If I select items in dimsum, should that selection be reflected when i click the 'TAT Trend' button?
- "/tat-trend?PIPELINE=kjfdslkjfds" just shows an empty plot. Probably alert on bad options

src/main/resources/templates/tat-trend.html
```diff
@@ -0,0 +1,63 @@
...
+            </select>
+          </div>
+        </div>
+        <div id="gatesCheckboxes" style="display: flex; gap: 10px; margin-top: 10px;">
+          <label class="font-inter font-medium text-12"><input type="checkbox" value="Receipt" checked> Receipt</label>
+          <label class="font-inter font-medium text-12"><input type="checkbox" value="Extraction" checked>
+            Extraction</label>
+          <label class="font-inter font-medium text-12"><input type="checkbox" value="Library Prep" checked> Library
+            Prep</label>
+          <label class="font-inter font-medium text-12"><input type="checkbox" value="Library Qual" checked> Library
```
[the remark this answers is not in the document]
wuall826: It is consistent with TAT report, so I think our users will be familiar

ts/tat-trend.ts
```diff
@@ -0,0 +1,437 @@
...
+    ((date.getTime() - onejan.getTime()) / 86400000 + onejan.getDay() + 1) / 7
+  );
+  return `${date.getFullYear()}-W${String(week).padStart(2, "0")}`;
+}
+
+function getColorByGate(): boolean {
+  const toggleColors = document.getElementById(
+    "toggleColors"
+  ) as HTMLInputElement;
+  return toggleColors ? toggleColors.checked : false;
```
djcooke, line 181: `toggleColors` should never be undefined here, right? Better to throw an error rather than handle an unexpected state in an unexpected way

ts/tat-trend.ts
```diff
@@ -0,0 +1,437 @@
...
+          : selectedGrouping === "fiscalQuarter"
+          ? "Fiscal Quarter"
+          : "Fiscal Year",
+      tickformat:
+        selectedGrouping === "week"
+          ? "%Y-W%U"
+          : selectedGrouping === "month"
+          ? "%Y-%m"
+          : selectedGrouping === "fiscalQuarter"
+          ? "%Y Q%q"
```
djcooke, line 258: Is Plotly determining the quarter based on this? Does that match up with ours?

ts/tat-trend.ts
```diff
@@ -0,0 +1,437 @@
...
+      selectedDataType
+    );
+    Plotly.newPlot("plotContainer", newPlot, layout);
+
+    window.addEventListener("resize", () => {
+      Plotly.relayout("plotContainer", {
+        width: window.innerWidth,
+        height: window.innerHeight,
+      });
+    });
```
djcooke, line 304: This can result in the plot being too small. Probably needs minimum dimensions

ts/tat-trend.ts
```diff
@@ -0,0 +1,437 @@
...
+
+function updatePlot(
+  selectedGrouping: string,
+  jsonData: any[],
+  selectedGates: string[],
+  selectedDataType: string
+) {
+  const trendReportContainer = document.getElementById("trendReportContainer");
+
+  if (trendReportContainer) {
```
djcooke, line 316: This being undefined should be an error, right? `html-utils` has a `getRequiredElementById` function for this

djcooke (commented):
Needs documentation added in `docs/user_manual/reports.md` (there's a place for it there already)

wuall826:
> Needs documentation added in `docs/user_manual/reports.md` (there's a place for it there already)

missed this, adding another commit

ts/tat-trend.ts
```diff
@@ -0,0 +1,446 @@
...
+          : selectedGrouping === "fiscalQuarter"
+          ? "Fiscal Quarter"
+          : "Fiscal Year",
+      tickformat:
+        selectedGrouping === "week"
+          ? "%Y-W%U"
+          : selectedGrouping === "month"
+          ? "%Y-%m"
+          : selectedGrouping === "fiscalQuarter"
+          ? "%Y Q%q" // Ensure this matches fiscal quarter determination
```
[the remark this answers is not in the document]
wuall826: yes, it matches

ts/tat-trend.ts
```diff
@@ -96,10 +95,13 @@ function getCompletedDateAndDays(
...
-      completedDate = row[`${dataTypePrefix} Release Completed`]
-        ? new Date(row[`${dataTypePrefix} Release Completed`])
+      const completedKey =
+        `${dataTypePrefix}_COMPLETED` as keyof typeof COLUMN_NAMES;
+      const daysKey = `${dataTypePrefix}_DAYS` as keyof typeof COLUMN_NAMES;
+      completedDate = row[COLUMN_NAMES[completedKey]]
+        ? new Date(row[COLUMN_NAMES[completedKey]])
         : null;
-      days = row[`${dataTypePrefix} Total Days`] ?? 0;
+      days = row[COLUMN_NAMES[daysKey]] ?? 0;
```
djcooke, line 104: The column names are still being constructed here. Casting them as constant values doesn't help.

All strings that refer to column names from the Case TAT Report should be either direct references to the constants, or use a function to construct them. This way, it's all in one place and much easier to maintain. There should also be comments by the constants and any such functions to explain that they are used for referring to the Case TAT Report.

djcooke (approved):

avarsava (approved):

PR #228, opened 2024-08-22
[GLT-4245] TAT trend report legend for Color by Gate

wuall826:
Jira ticket or GitHub issue:

- [x] Includes a change file
- [ ] Updated tests (or n/a)

### DIFF
--- a/changes/fix_GLT-4245
+++ b/changes/fix_GLT-4245
@@ -1,1 +1,1 @@
-TAT trend report grouping logic
+TAT Trend Report color by gate feature to work correctly with time range grouping
--- a/changes/fix_GLT-4245
+++ b/changes/fix_GLT-4245
@@ -1,0 +1,1 @@
+TAT trend report grouping logic
--- a/src/main/resources/templates/tat-trend.html
+++ b/src/main/resources/templates/tat-trend.html
@@ -25,6 +25,9 @@
             Year</button>
           <div style="margin-left: auto; display: flex; align-items: center; gap: 10px;">
             <label class="font-inter font-medium text-12">
+              <button id="legendButton" style="display:none;"
+                class="noprint bg-green-200 rounded-md hover:ring-2 ring-offset-1 ring-green-200 text-white font-inter font-medium text-12 px-2 py-1"
+                onclick="toggleLegend()">Legend</button>
               <input type="checkbox" id="toggleColors"> Color by Gate
             </label>
             <select id="dataSelection"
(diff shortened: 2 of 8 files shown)


ts/component/legend.ts
```diff
@@ -2,9 +2,22 @@ import { qcStatuses } from "../data/qc-status";
 import { makeIcon } from "../util/html-utils";
 
 const legendId = "legend-container";
+
+const GATE_COLORS = {
```
grapigeau, line 6: Why are some of the key is quotes ("Library Prep") and some not (Receipt)?

djcooke (commented):
The legend appears/disappears as intended, but this error is also showing up every time I click the Legend button:
image

changes/fix_GLT-4245
```diff
@@ -0,0 +1 @@
+TAT trend report grouping logic
```
djcooke, line 1: I'm not sure what this means? What feature was broken and is now fixed? Colour by gate?

src/main/resources/templates/tat-trend.html
```diff
@@ -11,52 +11,49 @@
...
+            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex items-center">Month</button>
           <button id="quarterButton" data-grouping="fiscalQuarter"
-            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex space-x-2 items-center">Fiscal
+            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex items-center">Fiscal
             Quarter</button>
           <button id="yearButton" data-grouping="fiscalYear"
-            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex space-x-2 items-center">Fiscal
+            class="font-inter font-medium text-12 text-black bg-grey-100 px-2 py-1 rounded-md hover:ring-2 ring-green-200 ring-offset-1 flex items-center">Fiscal
             Year</button>
```
djcooke, line 25: Remove `flex` and `items-center` from the buttons. These are flexbox (container) properties, and the buttons are flex items, so these are doing nothing.

ts/tat-trend.ts
```diff
@@ -403,6 +417,8 @@ function getSelectedDataType(): string {
 }
 
 document.addEventListener("DOMContentLoaded", () => {
+  const plotContainer = document.getElementById("plotContainer");
+  if (!plotContainer) return;
```
djcooke, line 421: What is the purpose of this check? Should it be an error?

wuall826: if not it throws an uncaught error on the QC report page `Uncaught Error: Required element weekButton not found` for all the buttons and dataSelection elements. It doesn't interfere with any rendering but I just wanted to get rid of the error.

djcooke: So that means `DOMContentLoaded` is firing multiple times, and sometimes it's before the DOM content is loaded..? I'm not sure why that would be, but it seems this is the wrong event to use. Would `load` work better?

wuall826: load did not work, but I can add the checks for elements

djcooke (approved):

grapigeau (approved):

PR #232, opened 2024-10-11
[GLT-4261] TAT trend report table metrics

wuall826:
Jira ticket or GitHub issue:

- [x] Includes a change file
- [ ] Updated tests (or n/a)

### DIFF
--- a/ts/component/table-builder.ts
+++ b/ts/component/table-builder.ts
@@ -694,22 +694,29 @@ export class TableBuilder<ParentType, ChildType> {
     if (this.definition.parentHeaders) {
       // create the first row with parent headers
       const parentRow = this.thead.insertRow();
+      if (this.definition.bulkActions) {
+        this.addSelectAllHeader(parentRow);
+      }
       this.definition.parentHeaders.forEach((parentHeader) => {
-        const th = document.createElement("th");
-        th.textContent = parentHeader.title;
-        th.colSpan = parentHeader.colspan; // set the colspan to cover subcolumns
-        th.className =
-          "p-4 font-semibold bg-grey-300 text-center text-white border-l-1 border-grey-200";
-        parentRow.appendChild(th);
+        addColumnHeader(
+          parentRow,
+          parentHeader.title,
+          false,
+          undefined,
+          "bg-grey-300",
+          parentHeader.colspan
+        );
       });
       // create the second row for child headers
       const childRow = this.thead.insertRow();
       this.columns.forEach((column, i) => {
-        const th = document.createElement("th");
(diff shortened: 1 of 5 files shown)


wuall826:
still need to stick "Assay" and "Gate" column, will add another commit. also, would it be better if I use the backend java csv download (like for other reports) for the metric download in this?

ts/tat-trend.ts
```diff
@@ -371,6 +433,174 @@ function updatePlotWithLegend(
...
+    { title: "", colspan: 1 },
+    ...sortedTimeRanges.map((timeRange) => ({
+      title: timeRange,
+      colspan: 3,
+    })),
+  ];
+  const caseTatTableDefinition: TableDefinition<AssayMetrics, void> = {
+    generateColumns: () => dynamicColumns,
+    parentHeaders, // pass parent headers for multi-level header support
+    disablePageControls: true,
```
djcooke, line 462: You've set sorting options for a couple columns, but this removes the sorting function

ts/tat-trend.ts
```diff
@@ -371,6 +433,174 @@ function updatePlotWithLegend(
...
+    filters: [],
+    staticActions: [],
```
djcooke, line 464: These are both optional. Exclude them rather than assigning an empty array

avarsava (approved):

ts/tat-trend.ts
```diff
@@ -597,13 +601,22 @@ function buildTableFromMetrics(
...
           assayMetrics.timeRanges.push({
             group: timeRange,
-            avgDays: averageDays,
-            medianDays,
+            avgDays:
+              averageDays !== undefined
+                ? averageDays.toString()
+                : NOT_AVAILABLE,
+            medianDays:
+              medianDays !== undefined ? medianDays.toString() : NOT_AVAILABLE,
```
djcooke, line 619: You're setting these values to undefined and then changing undefined to N/A right below. Either store as undefined in the time range (and handle that appropriately elsewhere), or set these to N/A initially.

ts/component/table-builder.ts
```diff
@@ -708,29 +718,33 @@ export class TableBuilder<ParentType, ChildType> {
...
-        this.selectAllCheckbox = undefined;
       }
       this.columns.forEach((column, i) => {
         addColumnHeader(
           row,
           column.title,
-          i == 0 && !this.definition.bulkActions,
+          i === 0 && !this.definition.bulkActions,
           column.headingClass
         );
```
djcooke, line 749: This is all duplicated. The only difference is the last two parameters of `addColumnHeader`

djcooke (approved):

PR #242, opened 2024-12-05
[GLT-4307] Added donor assay report

wuall826:
Jira ticket or GitHub issue:

- [ ] Includes a change file
- [ ] Updated tests (or n/a)

### DIFF
--- a/src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
@@ -1,0 +1,89 @@
+package ca.on.oicr.gsi.dimsum.util.reporting.reports;
+
+import java.time.LocalDate;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+import com.fasterxml.jackson.databind.JsonNode;
+import ca.on.oicr.gsi.cardea.data.Case;
+import ca.on.oicr.gsi.cardea.data.Project;
+import ca.on.oicr.gsi.dimsum.controller.BadRequestException;
+import ca.on.oicr.gsi.dimsum.service.CaseService;
+import ca.on.oicr.gsi.dimsum.service.filtering.CompletedGate;
+import ca.on.oicr.gsi.dimsum.util.reporting.Column;
+import ca.on.oicr.gsi.dimsum.util.reporting.Report;
+import ca.on.oicr.gsi.dimsum.util.reporting.ReportSection;
+import ca.on.oicr.gsi.dimsum.util.reporting.ReportSection.TableReportSection;
+import ca.on.oicr.gsi.dimsum.util.reporting.reports.shared.CaseSampleRowData;
+
+public class DonorAssayReport extends Report {
+
+  private static final ReportSection<CaseSampleRowData> mainSection =
+      new TableReportSection<CaseSampleRowData>("Donor Assay Report", Arrays.asList(
+          Column.forString("Project", x -> x.getSample().getProject()),
+          Column.forString("Pipeline", x -> getProjectPipelines(x.getCase())),
+          Column.forString("Donor", x -> x.getCase().getDonor().getName()),
+          Column.forString("External Name", x -> x.getCase().getDonor().getExternalName()),
+          Column.forString("Assay", x -> x.getCase().getAssayName()),
+          Column.forString("Start Date", x -> getStartDate(x.getCase()).toString()),
+          Column.forString("Latest Activity", x -> getLatestActivityDate(x.getCase()).toString()),
+          Column.forString("Status", x -> getCaseStatus(x.getCase())),
+          Column.forString("Stop/Pause Reason", x -> getStopPauseReason(x.getCase())))) {
+        @Override
+        public List<CaseSampleRowData> getData(CaseService caseService, JsonNode parameters) {
+          Set<String> caseIds = getParameterStringSet(parameters, "caseIds");
+          if (caseIds == null) {
+            throw new BadRequestException("caseIds parameter missing");
+          }
+          return CaseSampleRowData.listByCaseIds(caseService, caseIds);
+        }
+      };
+
+  public static final DonorAssayReport INSTANCE = new DonorAssayReport();
+
+  private DonorAssayReport() {
+    super("Donor Assay Report", mainSection);
+  }
+
+  private static String getProjectPipelines(Case kase) {
+    return kase.getProjects().stream()
+        .map(Project::getPipeline)
+        .filter(Objects::nonNull)
+        .collect(Collectors.joining(", "));
+  }
+
+  private static LocalDate getStartDate(Case kase) {
+    return kase.getStartDate();
+  }
+
+  private static LocalDate getLatestActivityDate(Case kase) {
+    return kase.getLatestActivityDate();
+  }
+
+  private static String getCaseStatus(Case kase) {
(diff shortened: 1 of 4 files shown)


avarsava (approved):

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
```diff
@@ -0,0 +1,89 @@
...
+          Column.forString("Latest Activity", x -> getLatestActivityDate(x.getCase()).toString()),
+          Column.forString("Status", x -> getCaseStatus(x.getCase())),
+          Column.forString("Stop/Pause Reason", x -> getStopPauseReason(x.getCase())))) {
+        @Override
+        public List<CaseSampleRowData> getData(CaseService caseService, JsonNode parameters) {
+          Set<String> caseIds = getParameterStringSet(parameters, "caseIds");
+          if (caseIds == null) {
+            throw new BadRequestException("caseIds parameter missing");
+          }
+          return CaseSampleRowData.listByCaseIds(caseService, caseIds);
```
djcooke, line 40: Shouldn't this report include one row per case? This creates one row per full-depth sequencing run-library, and if there are none will exclude the case.

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
```diff
@@ -0,0 +1,89 @@
...
+import ca.on.oicr.gsi.dimsum.util.reporting.ReportSection;
+import ca.on.oicr.gsi.dimsum.util.reporting.ReportSection.TableReportSection;
+import ca.on.oicr.gsi.dimsum.util.reporting.reports.shared.CaseSampleRowData;
+
+public class DonorAssayReport extends Report {
+
+  private static final ReportSection<CaseSampleRowData> mainSection =
+      new TableReportSection<CaseSampleRowData>("Donor Assay Report", Arrays.asList(
+          Column.forString("Project", x -> x.getSample().getProject()),
+          Column.forString("Pipeline", x -> getProjectPipelines(x.getCase())),
```
djcooke, line 26: This is correctly including the pipelines from all of the case's projects, but isn't showing all of the case's projects

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
```diff
@@ -0,0 +1,89 @@
...
+        .collect(Collectors.joining(", "));
+  }
+
+  private static LocalDate getStartDate(Case kase) {
+    return kase.getStartDate();
+  }
+
+  private static LocalDate getLatestActivityDate(Case kase) {
+    return kase.getLatestActivityDate();
+  }
```
djcooke, line 63: These are not useful - just turning `x.getStartDate()` into `getStartDate(x)`

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
```diff
@@ -0,0 +1,89 @@
...
+    if (kase.isStopped()) {
+      return kase.getRequisition().getStopReason() != null
+          ? kase.getRequisition().getStopReason()
+          : "Reason Unspecified";
+    } else if (kase.getRequisition().isPaused()) {
+      return kase.getRequisition().getPauseReason() != null
+          ? kase.getRequisition().getPauseReason()
+          : "Reason Unspecified";
+    }
+    return "";
```
djcooke, line 87: No need for empty Strings. Can return null

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
```diff
@@ -1,43 +1,65 @@
...
+          if (caseIds == null || caseIds.isEmpty()) {
+            throw new BadRequestException("caseIds parameter missing or empty");
           }
-          return CaseSampleRowData.listByCaseIds(caseService, caseIds);
+          List<Case> cases = caseService.getCasesByIds(caseIds);
+          Map<Long, Assay> assaysById = caseService.getAssaysById();
+          return cases.stream()
+              .flatMap(kase -> kase.getTests().stream()
+                  .map(test -> new RowData(kase, test, assaysById.get(kase.getAssayId()))))
+              .collect(Collectors.toList());
```
djcooke, line 62: Now the report will create duplicate rows for every test in a case, and you're not even using the test or assay that you feed to the RowData constructor. No transformation is necessary, and the RowData class is not necessary. This should simply return the list of cases.

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
```diff
@@ -0,0 +1,81 @@
...
+        .map(Project::getPipeline)
+        .filter(Objects::nonNull)
+        .collect(Collectors.joining(", "));
+  }
+
+  private static String getCaseStatus(Case kase) {
+    if (CompletedGate.RELEASE.qualifyCase(kase)) {
+      return "Completed";
+    } else if (kase.isStopped()) {
+      return "Failed";
```
djcooke, line 61: 1. Text for stopped should be "Stopped" rather than failed - we don't know why it was stopped, and it might not be a failure. After the new release QCs are out, there can also be failed cases that are not stopped, making "Failed" more confusing since it wouldn't be used for those.
2. I think the check for stopped should come before the check for completion. A stopped case can (and should) also be completed and I think a stopped+completed case should show as stopped. Might want to check with Morgan what she would prefer though.

wuall826: Morgan said stopped is enough for now

wuall826: going with Mike's 2 column suggestion

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
```diff
@@ -0,0 +1,92 @@
...
+          Column.forString("Project", kase -> kase.getProjects().stream()
+              .map(Project::getName)
+              .collect(Collectors.joining(", "))),
+          Column.forString("Pipeline", DonorAssayReport::getProjectPipelines),
+          Column.forString("Donor", kase -> kase.getDonor().getName()),
+          Column.forString("External Name", kase -> kase.getDonor().getExternalName()),
+          Column.forString("Assay", Case::getAssayName),
+          Column.forString("Start Date", kase -> kase.getStartDate().toString()),
+          Column.forString("Latest Activity", kase -> kase.getLatestActivityDate().toString()),
+          Column.forString("Status", DonorAssayReport::getCaseStatus),
```
djcooke, line 32: Remove this column since it's been replaced by "Completed" and "Stopped/Paused"

src/main/java/ca/on/oicr/gsi/dimsum/util/reporting/reports/DonorAssayReport.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private static String getCaseStatus(Case kase) {
+    if (CompletedGate.RELEASE.qualifyCase(kase)) {
+      return "Completed";
+    } else if (kase.getRequisition().isPaused()) {
+      return "Paused";
+    } else {
+      return "In Progress";
+    }
+  }
```
djcooke, line 67: Unused

djcooke (approved):

PR #249, opened 2024-12-23
[GLT-4314] Simplified SAML2 internal configuration

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4314

- [x] Updated tests (or n/a)

This would probably work with our current production IdP, but there's no good way to test that and no advantage to merging it into main right now anyway. We're going to need a branch for external-dimsum at least until we're ready to switch over to Keycloak and role-based access, so I've started the `dimsum-external` branch for that.

If you'd like to test using keycloak-dev, let me know and I can help you set it up. You'll need a login for keycloak-dev with realm admin permissions, and you'll have to run Dimsum on a URL besides "http://localhost:8081" since I am using that. This doesn't need to be a URL that's visible to Keycloak, so localhost works with a different port, or you can create yourself an alias in `/etc/hosts`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 2. Copy [example-application.properties](example-application.properties) into the `config`
    directory and rename it to `application.properties`
 3. Are you enabling SAML authentication?
-   * If yes:
+   - If yes:
      1. add the IdP certificate to your `config` directory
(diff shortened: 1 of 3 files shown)


wuall826 (approved):

callunity (approved):
That's a lot of beans removed!

PR #250, opened 2025-01-06
[GLT-4316] created DTOs for external view

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4316

- [x] Updated tests (or n/a)

### DIFF
--- a/src/main/java/ca/on/oicr/gsi/dimsum/data/external/ExternalAssay.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/data/external/ExternalAssay.java
@@ -1,0 +1,35 @@
+package ca.on.oicr.gsi.dimsum.data.external;
+
+import ca.on.oicr.gsi.cardea.data.Assay;
+
+public class ExternalAssay {
+
+  private final long id;
+  private final String name;
+  private final String version;
+  private final String description;
+
+  public ExternalAssay(Assay from) {
+    id = from.getId();
+    name = from.getName();
+    version = from.getVersion();
+    description = from.getDescription();
+  }
+
+  public long getId() {
+    return id;
+  }
+
+  public String getName() {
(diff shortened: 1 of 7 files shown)


src/main/java/ca/on/oicr/gsi/dimsum/data/external/ExternalAssay.java
```diff
@@ -0,0 +1,35 @@
+package ca.on.oicr.gsi.dimsum.data.external;
+
+import ca.on.oicr.gsi.cardea.data.Assay;
+
+public class ExternalAssay {
```
callunity, line 5: Would this be a good scenario for using Java `record` ? This class could be simplified to:
```
public record ExternalAssay  (long id, String name, String version, String description) {
  public ExternalAssay(Assay from) {
    this(from.getId(), from.getName(), from.getVersion(), from.getDescription());
  }
}
```
The constructors for the ExternalCase and some of the others would be a bit more involved, and would require some comments about e.g. `internal models x, y, z have no sensitive fields`, but the guarantees of immutability and built-in equals, hashCode and toString are nice.

callunity (approved):

wuall826 (approved):

src/main/java/ca/on/oicr/gsi/dimsum/data/external/ExternalAssay.java
djcooke: I thought the constructors were pretty gross when there were a lot of fields, but you're right - they're only used once inside the class, and the benefits easily outweigh that. Thanks for convincing me!

djcooke:
Changed all to records and also added `latestActivityDate` and `nucleicAcid` type back into `ExternalSample`

PR #253, opened 2025-01-10
[GLT-4319] adapted main dashboard for external users

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4319

- [x] Includes a change file
- [x] Updated tests (or n/a)

### DIFF
--- a/ts/data/case.ts
+++ b/ts/data/case.ts
@@ -19,6 +19,7 @@ import {
   getMetricCategory,
   getReleaseApprovalQcStatus,
   getReleaseQcStatus,
+  internalUser,
   siteConfig,
 } from "../util/site-config";
 import {
@@ -222,20 +223,24 @@ export const caseDefinition: TableDefinition<Case, Test> = {
     {
       title: "TAT Report",
       handler: showTatReportDialog,
+      view: "internal",
     },
     {
       title: "TAT Trend",
       handler: showTatTrendPage,
+      view: "internal",
     },
   ],
   bulkActions: [
     {
       title: "Download",
       handler: showDownloadDialog,
+      view: "internal",
     },
     {
       title: "Sign Off",
       handler: showSignoffDialog,
+      view: "internal",
     },
   ],
   generateColumns: () => [
@@ -362,9 +367,11 @@ export const caseDefinition: TableDefinition<Case, Test> = {
           addNaText(fragment);
           return;
         }
-        addSampleIcons(kase.assayId, test.extractions, fragment, true);
+        addSampleIcons(kase.assayId, test.extractions, fragment, internalUser);
         if (samplePhaseComplete(kase.receipts)) {
-          if (samplePhasePendingWorkQcOrTransfer(test.extractions, true)) {
+          if (
+            samplePhasePendingWorkQcOrTransfer(test.extractions, internalUser)
+          ) {
             if (
               samplePhasePendingWork(test.extractions) &&
               !kase.requisition.paused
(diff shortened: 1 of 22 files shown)


ts/data/case.ts
```diff
@@ -1494,16 +1505,18 @@ export function makeSampleTooltip(sample: Sample) {
...
-            : sample.run.name,
-          urls.miso.run(sample.run.name),
-          urls.dimsum.run(sample.run.name),
-          sample.run.name
-        )
+      const runNameContainer = makeNameDiv(
+        sample.sequencingLane
+          ? sample.run.name + " (L" + sample.sequencingLane + ")"
+          : sample.run.name,
+        urls.miso.run(sample.run.name),
```
callunity, line 1512: Do external users have access to MISO?

djcooke: No. The `makeNameDiv` function only creates a link for internal users

callunity: Ah - that's a bit unclear, since the next line down has different behaviour depending on whether the user is internal (which is why I was wondering why there didn't seem to be a similar internalUser conditional for the MISO link). Is this because some dimsum URLs can be displayed to internal users, and others shouldn't?

djcooke: Yeah, exactly. External users will be able to access donor, project, and requisition detail pages, but won't have access to the run details page.

There is a lot of inconsistency in how internal details are controlled - some are simply not shown if the relevant data field is undefined, some explicitly have an `internalUser` check, and some are controlled within helper methods like `makeNameDiv`. The alternative is having tons of `internalUser` checks everywhere, which I think makes more of a mess. I tried to keep the front-end changes minimal.

What's most important is that there's no chance of a front-end bug displaying something that external users shouldn't see because no data that they shouldn't have should ever be sent to the front end. Similarly, if we did accidentally create a link to the run page, it would be a broken link and possibly cause confusion, but it would never allow the user to access run details that they aren't authorized to see.

callunity: I do worry that since we don't have automated tests for our display, someone could accidentally make a change (e.g. to the makeDiv function) while you're on vacation that could have some unexpected and far-reaching effects. As you said, those effects would be limited to broken links, but these would be only visible to external users, which has more negative reputational effects on us than if it was only visible to internal users. I'd personally prefer to have explicit and messy `internalUser` checks if it reduces the likelihood of hard-to-catch errors, but perhaps we can figure this out with processes instead, or something else. 

Do you have credentials for an external user that we can use for testing during development and PR reviews?

djcooke: I don't think adding more explicit checks reduces the likelihood of errors because that doesn't prevent someone else from forgetting or not realizing they need a similar check somewhere else.

Marking fields as optional in the typescript interfaces forces the user to deal with fields that may be undefined, which does help prevent errors, and I don't think we should put `internalUser` checks on top of that. `internalUser` checks are only necessary where that doesn't suffice, such as for run links - the run name is there and can be displayed, but should not have a link.

To test the external view yourself, you would currently need to set up your Dimsum to use Keycloak, which I can help you with if you'd like. We could add options to switch the user to external and specify their projects when authentication is disabled. I'll create a separate ticket for that.

callunity: Yes please to help with setting up Dimsum to use Keycloak, and +1 to the separate ticket for easier development.

wuall826 (approved):

callunity (approved):

PR #257, opened 2025-01-28
[GLT-4337] added more multi-project case info to FAQ

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4337

- [x] Includes a change file
- [x] Updated tests (or n/a)

### DIFF
--- a/docs/user_manual/faq.md
+++ b/docs/user_manual/faq.md
@@ -98,6 +98,44 @@ own project
 *You can hover over items in Dimsum to see their project*
 
 
+## How do I limit my case to a single project?
+
+Some cases are associated with multiple projects. This is usually caused by samples having been
+previously received into one project, then the samples themselves were used for sequencing in a
+different project. If the samples had previous data that could be used, then they could have been
+added as supplemental to the new requisition, and supplemental samples do not effect the case's
+projects. If the samples had to be derived and sequenced, they should have undergone the process
+described in the example below instead:
+
+X1 is a tumour tissue sample – received in Project A for WTS
+
+- Accessioned into Project A; Requisition 1 – WTS Assay
+- Dual extraction done: X1TDNA and X1TRNA are the propagated samples
+- X1TRNA aliquot transferred to library prep team
+
+X2 is a buffy coat sample – received for Project B a year later for WGTS
+
+- Accessioned into Project A, then changed to Project B (assumes knowledge that this is for a
+  different project that already exists – often true but not always); Requisition 2 – WGTS Assay
+- DNA extraction done: X2BCDNA is the propagated sample
+- X2BCDNA aliquot transferred to library prep team
+- X1TDNA aliquot transferred to library prep team
+
+Option 1:
+
+X1TDNA stock and X1TRNA aliquot can be added as a supplemental samples to Req 2. This needs to be
+the case with X1TRNA if the same WTS data previously generated is being used for the newly requested
+WGTS Assay.
+
+Option 2:
+
+X1TDNA and X1TRNA should have new aliquots propagated into Project B. These should be added as
+requisitioned samples to Requisition 2.
+
+Using either option, the requisitioned samples in Requisition 2 and everything downstream of them
+are associated with Project B, so that will be the only project associated with the WGTS case.
+
+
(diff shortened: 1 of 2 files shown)


docs/user_manual/faq.md
```diff
@@ -98,6 +98,44 @@ own project
...
+- Dual extraction done: X1TDNA and X1TRNA are the propagated samples
+- X1TRNA aliquot transferred to library prep team
+
+X2 is a buffy coat sample – received for Project B a year later for WGTS
+
+- Accessioned into Project A, then changed to Project B (assumes knowledge that this is for a
+  different project that already exists – often true but not always); Requisition 2 – WGTS Assay
+- DNA extraction done: X2BCDNA is the propagated sample
+- X2BCDNA aliquot transferred to library prep team
+- X1TDNA aliquot transferred to library prep team
```
callunity, line 122: What do you think about `- X1TDNA aliquot (from tumour tissue sample extraction above) transferred to library prep team` for clarify?

docs/user_manual/faq.md
```diff
@@ -98,6 +98,44 @@ own project
...
+Option 1:
+
+X1TDNA stock and X1TRNA aliquot can be added as a supplemental samples to Req 2. This needs to be
+the case with X1TRNA if the same WTS data previously generated is being used for the newly requested
+WGTS Assay.
+
+Option 2:
+
+X1TDNA and X1TRNA should have new aliquots propagated into Project B. These should be added as
+requisitioned samples to Requisition 2.
```
callunity, line 133: If this is the scenario used, would `X1TRNA` be required to be sequenced again for req 2?

docs/user_manual/faq.md
```diff
@@ -98,6 +98,44 @@ own project
...
+  different project that already exists – often true but not always); Requisition 2 – WGTS Assay
+- DNA extraction done: X2BCDNA is the propagated sample
+- X2BCDNA aliquot transferred to library prep team
+- X1TDNA aliquot transferred to library prep team
+
+Option 1:
+
+X1TDNA stock and X1TRNA aliquot can be added as a supplemental samples to Req 2. This needs to be
+the case with X1TRNA if the same WTS data previously generated is being used for the newly requested
+WGTS Assay.
```
callunity, line 128: I'd add a line break between the two, or reword to:

If the WTS data previously generated from X1TRNA is being used for the newly-requested WGTS Assay, X1TRNA aliquot must be added as a supplemental sample to Req 2. 

X1TDNA stock can also be added as a supplemental sample to Req 2.

docs/user_manual/faq.md
```diff
@@ -98,6 +98,44 @@ own project
...
+
+- Accessioned into Project A, then changed to Project B (assumes knowledge that this is for a
+  different project that already exists – often true but not always); Requisition 2 – WGTS Assay
+- DNA extraction done: X2BCDNA is the propagated sample
+- X2BCDNA aliquot transferred to library prep team
+- X1TDNA aliquot transferred to library prep team
+
+Option 1:
+
+X1TDNA stock and X1TRNA aliquot can be added as a supplemental samples to Req 2. This needs to be
```
avarsava, line 126: "a supplemental samples" -> "supplemental samples"

avarsava (approved):

docs/user_manual/faq.md
djcooke: Adding "(extraction from Requisition 1)" to clarify

docs/user_manual/faq.md
djcooke: Yes. Clarifying

docs/user_manual/faq.md
djcooke: Actually, it's more complicated... There's no reason to add these separately as supplemental if you want them both - you could just add the tissue as supplemental. The problem either way is that X1TDNA isn't actually used in the original req's case (assay doesn't use DNA), so it doesn't have any metrics. It would show up in the second case as supplemental, but without metrics. Rewriting some of this to clarify the effects

callunity (approved):

PR #278, opened 2025-07-03
[GLT-4425] fixed 'QC in MISO' to omit pass/fail metrics

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4425

- [x] Includes a change file
- [x] Updated tests (or n/a)


avarsava:
Doesn't this mean something could fail and it'd be QC'd in MISO as passed?

PR #283, opened 2025-08-11
[GLT-4348] fail authentication if user full name is not found

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4348

- [x] Updated tests (or n/a)


djcooke:
> Mononymous people not supported?

I don't know how we should handle that based on the fields in Keycloak, but I also don't anticipate this being a requirement.

djcooke:
If first name or last name is missing, it could be because it was intentionally omitted (mononymous person), accidentally omitted, or misconfiguration. I think accidental omission and misconfiguration are both much more likely to happen than intentional omission, and would affect what is recorded for sign-offs. If we ever do have a mononymous user, it would not be difficult to update this to support that, and maybe we should do that before (if) it happens. At least initially, I think it's much safer to expect both for now though.

PR #287, opened 2025-09-15
added external user manual

djcooke:
Jira ticket or GitHub issue: n/a

- [x] Includes a change file
- [x] Updated tests (or n/a)


src/main/resources/templates/layout.html
```diff
@@ -44,8 +44,8 @@
         <a th:href="${bugReportUrl}" target="_blank" rel="noopener noreferrer"
           class="font-inter text-16 text-black hover:text-green-200 float-right pl-4 inline-block mt-2">report
           a problem</a>
-        <a th:href="'https://dimsum-docs.readthedocs.io/en/' + ${docsVersion} + '/'" target="_blank"
-          rel="noopener noreferrer"
+        <a th:href="'https://dimsum-docs.readthedocs.io/en/' + ${docsVersion} + '/' + ${internalUser ? '' : 'user_manual/external/'}"
```
callunity, line 47: Why does the internal user manual not have namespacing on readthedocs?

djcooke: It does except for the index page. The site should have a root index. That could either be the existing one, or it could be a page that just links to both the internal and external manual ones. I chose the former and added a link to the external manual at the top. Open to other suggestions though.

PR #327, opened 2026-04-29
[GLT-4781] allow self-assigning releases

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4781

- [x] Includes a change file
- [x] Updated tests (or n/a)

### DIFF
--- a/src/main/java/ca/on/oicr/gsi/dimsum/data/CacheUpdatedCase.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/data/CacheUpdatedCase.java
@@ -35,7 +35,8 @@ public class CacheUpdatedCase implements Case {
       Map<String, String> categoryAssignments =
           releaseAssignments.get(original.getDeliverableCategory());
       if (categorySignoff != null || categoryAssignments != null) {
-        deliverables.add(new CacheUpdatedCaseDeliverable(original, signoff, categoryAssignments));
+        deliverables
+            .add(new CacheUpdatedCaseDeliverable(original, categorySignoff, categoryAssignments));
       } else {
         deliverables.add(original);
       }
--- a/src/main/java/ca/on/oicr/gsi/dimsum/data/CacheUpdatedCase.java
+++ b/src/main/java/ca/on/oicr/gsi/dimsum/data/CacheUpdatedCase.java
@@ -5,6 +5,7 @@ import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import ca.on.oicr.gsi.cardea.data.AnalysisQcGroup;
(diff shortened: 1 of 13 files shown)


src/main/java/ca/on/oicr/gsi/dimsum/data/CacheUpdatedCase.java
```diff
@@ -22,13 +23,19 @@ public class CacheUpdatedCase implements Case {
...
-      if (Objects.equals(original.getDeliverableCategory(), signoff.getDeliverableType())) {
-        deliverables.add(new CacheUpdatedCaseDeliverable(original, signoff));
+      NabuSavedSignoff categorySignoff = signoff != null
+          && Objects.equals(original.getDeliverableCategory(), signoff.getDeliverableType())
+              ? signoff
+              : null;
+      Map<String, String> categoryAssignments =
+          releaseAssignments.get(original.getDeliverableCategory());
+      if (categorySignoff != null || categoryAssignments != null) {
+        deliverables.add(new CacheUpdatedCaseDeliverable(original, signoff, categoryAssignments));
```
avarsava: If signoff is not null but categorySignoff is null (therefore the Objects.equals failed), then categoryAssignments contains releaseAssignments for a deliverable category other than signoff, but they are added together?

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -1089,10 +1240,99 @@ private void removeExpiredCachedSignoffs() {
...
+          if (categoryAssignments.getValue().isEmpty()) {
+            categoryIterator.remove();
+          }
+        }
+        if (caseAssignments.getValue().isEmpty()) {
+          caseIterator.remove();
+        }
+      }
+    }
+    tryDumpAssignments();
```
avarsava: Should this also be synchronized on cachedSignoffsByCaseId? Someone could theoretically be causing `assignmentsChanged` to flip at midnight

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
```diff
@@ -1058,21 +1102,128 @@ public void cacheSignoffs(Collection<NabuSavedSignoff> signoffs) {
...
+    synchronized (cachedSignoffsByCaseId) {
+      Map<String, Map<String, Map<String, String>>> previousAssignmentsByCaseId = new HashMap<>();
+      for (NabuBulkSignoff signoff : signoffs) {
+        for (String caseId : signoff.getCaseIdentifiers()) {
+          Map<String, Map<String, String>> caseAssignments =
+              cachedReleaseAssignments.computeIfAbsent(caseId, (x) -> new HashMap<>());
+          Map<String, String> categoryAssignments =
+              caseAssignments.computeIfAbsent(signoff.getDeliverableType(), (x) -> new HashMap<>());
+
+          // remember previous value in-case we need to cancel
```
avarsava, line 1135: Would it be worth bundling these into a cloneable class to keep as a memento?

avarsava (commented):
Should this change include updates to the user manual?

src/main/java/ca/on/oicr/gsi/dimsum/data/CacheUpdatedCase.java
djcooke: `CacheUpdatedCase` holds a `baseCase` and functions mostly as a wrapper for it. A cached signoff that the `CacheUpdatedCase` was built for will be returned where appropriate, and data from `baseCase` is used otherwise. `baseCase` itself may be a `CacheUpdatedCase` so it may be returning another cached signoff too.

`baseCase` is a Cardea `Case` though, which doesn't have assignment. I'm including the assignments in every build of `CacheUpdatedCase` to get around this because a lot of gross `instanceof`/casting would be needed otherwise.

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
djcooke: Yes, definitely. Thanks for catching that it was outside of the synchronized block!

src/main/java/ca/on/oicr/gsi/dimsum/service/CaseService.java
djcooke: I considered it, but we'd still need the maps elsewhere for efficiency, and cloning/remembering previous values is only needed here, so not really worth it.

src/main/java/ca/on/oicr/gsi/dimsum/data/CacheUpdatedCase.java
avarsava: Sorry, I still don't understand. If signoff is of deliverable "type" (which appears to `equal` to deliverable category?) Y and original is of deliverable category X, then categorySignoff will be null, but categoryAssignments will still get the releaseAssignments of type X. It therefore looks like on line 38 there is a possibility of adding `new CacheUpdatedCaseDeliverable(original of deliverable category X, signoff of deliverable category Y, categoryAssignments of deliverable category X)` . This is OK?

djcooke: Ah, you're right, thanks! As is, if there was a deliverable named the same for two different categories, assigning one would affect them both. Will change `signoff` -> `categorySignoff` to fix

grapigeau (approved):

avarsava (approved):

PR #336, opened 2026-07-23
add run level metrics to  samples

grapigeau:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GP-5765

- [x] Includes a change file
- [ ] Updated tests (or n/a)

I think this minimal change is looking good but let me know if I missed something

### DIFF
--- a/changes/add_run_level_metrics.md
+++ b/changes/add_run_level_metrics.md
@@ -1,1 +1,0 @@
-Run Level metric added to  samples
--- a/changes/add_run_level_metrics.md
+++ b/changes/add_run_level_metrics.md
@@ -1,0 +1,1 @@
+Run Level metric added to  samples
--- a/ts/data/sample.ts
+++ b/ts/data/sample.ts
@@ -758,9 +758,11 @@ export function getSampleMetricCellHighlight(
     return null;
   }
 
-  // TODO: handle run and lane level metrics (not populated at time of writing)
+  // TODO: handle lane level metrics (not populated at time of writing)
   const sampleMetric = sample.metrics.find(
-    (metric) => metric.name === metricName && metric.metricLevel == "SAMPLE",
+    (metric) =>
+      metric.name === metricName &&
+      (metric.metricLevel == "SAMPLE" || metric.metricLevel == "RUN"),
   );
   // handle metrics that may be preliminary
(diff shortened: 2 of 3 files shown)


changes/add_run_level_metrics.md
```diff
@@ -0,0 +1 @@
+Run Level metric added to  samples
```
djcooke: It's unclear what this means, but this is not really a user-visible change, so doesn't need a change file

ts/data/sample.ts
```diff
@@ -758,9 +758,11 @@ export function getSampleMetricCellHighlight(
     return null;
   }
 
-  // TODO: handle run and lane level metrics (not populated at time of writing)
+  // TODO: handle lane level metrics (not populated at time of writing)
```
djcooke: Lane-level metrics are populated now, so you can remove that bit too

callunity (approved):

ts/data/sample.ts
```diff
@@ -869,9 +871,11 @@ export function addMetricValueContents(
     addTextDiv("Manual check required", fragment);
     return;
   }
-  // TODO: handle run and lane level metrics (not populated at time of writing)
+  // TODO: handle lane level metrics (not populated at time of writing)
```
djcooke: Again, lane-level metrics are populated. This doesn't support lane-level metrics though, and also doesn't support run metrics with multiple values (handled above, but not using the metrics collection), if you could update the TODO to reflect that please

djcooke (commented):
This doesn't handle the run-level Illumina metrics since they have multiple values, but it works for the new Ultima metrics, which was your goal, so that's fine.

The commit message is confusing though since samples already had run-level metrics. Please also include the ticket number.

ts/data/sample.ts
grapigeau: Sorry, a bit confused by this message?

djcooke:
```suggestion
  // TODO: handle lane level metrics and run level metrics with multiple values
```

grapigeau: Ok ok, I was confused by the wording between "populated" vs "supported". Makes sense, thanks

djcooke (approved):

